PR #2, opened 2019-04-05
fix: Create gitignore for node_modules and other files

### LINKED ISSUE #1: No .gitignore file
This project does not have a `.gitignore`. This will cause trouble developers contributing to the repo who wish to add files using `git add -A`. I assume not all developers have a global `.gitignore`.

I suggest using a `.gitignore` file like the Node client:
https://github.com/googleapis/google-api-nodejs-client/blob/master/.gitignore

grant:
Creates a `.gitignore` file.
Fixes #1.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,5 @@ npm-debug.log
 .vscode
 .idea
 .DS_Store
-package-lock.json
 yarn.lock
+yarn-error.log
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,11 @@
+node_modules/
+build/
+/docs/
+.coverage/
+npm-debug.log
+.nyc_output
+.vscode
+.idea
+.DS_Store
+package-lock.json
+yarn.lock


.gitignore
```diff
@@ -0,0 +1,11 @@
+node_modules/
+build/
+/docs/
+.coverage/
+npm-debug.log
+.nyc_output
+.vscode
+.idea
+.DS_Store
+package-lock.json
```
swalkowski: We want package-lock.json to be checked in. I don't think it should be ignored then?

.gitignore
```diff
@@ -0,0 +1,11 @@
...
+yarn.lock
```
swalkowski: Let's ignore yarn-error.log as well.

grant: Done.

.gitignore
grant: Correct. Reverted.

PR #16, opened 2019-04-11
fix #4: Removes 'function' prefix from CLI args

### LINKED ISSUE #4: Flags use an unnecessary `function-` prefix
all flags are prefixed with `--function-`

The "function" context is already captured in the framework name.

I recommend going for: `functions-framework --target [TARGET] --signature-type [Type]`

grant:
Fixes #4.
- Removes redundant `function` from CLI/env vars for `target` and `signature-type`
- Updates docs
- Moves string literals into constants to reduce renaming multiple pieces of code and clearly define the library interface in 1 place.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -31,7 +31,7 @@ import * as minimist from 'minimist';
 
 import {
   ErrorHandler,
-  FunctionSignatureType,
+  SignatureType,
   getServer,
   getUserFunction,
 } from './invoker';
@@ -46,8 +46,8 @@ const FLAG = {
 // Supported environment variables
 const ENV = {
   PORT: 'PORT',
-  TARGET: 'TARGET',
-  SIGNATURE_TYPE: 'SIGNATURE_TYPE', // underscore
+  TARGET: 'FUNCTION_TARGET',
+  SIGNATURE_TYPE: 'FUNCTION_SIGNATURE_TYPE', // underscore
 };
 
 enum NodeEnv {
@@ -62,13 +62,13 @@ const CODE_LOCATION = process.cwd();
 const PORT = argv[FLAG.PORT] || process.env[ENV.PORT] || '8080';
 const TARGET = argv[FLAG.TARGET] || process.env[ENV.TARGET] || 'function';
 
-const FUNCTION_SIGNATURE_TYPE_STRING =
+const SIGNATURE_TYPE_STRING =
   argv[FLAG.SIGNATURE_TYPE] || process.env[ENV.SIGNATURE_TYPE] || 'http';
-const FUNCTION_SIGNATURE_TYPE =
-  FunctionSignatureType[
-    FUNCTION_SIGNATURE_TYPE_STRING.toUpperCase() as keyof typeof FunctionSignatureType
+const SIGNATURE_TYPE =
+  SignatureType[
+    SIGNATURE_TYPE_STRING.toUpperCase() as keyof typeof SignatureType
   ];
-if (FUNCTION_SIGNATURE_TYPE === undefined) {
+if (SIGNATURE_TYPE === undefined) {
   console.error(`Function signature type must be one of 'http' or 'event'.`);
   process.exit(1);
 }
@@ -79,7 +79,7 @@ if (!USER_FUNCTION) {
   process.exit(1);
 }
 
-const SERVER = getServer(USER_FUNCTION!, FUNCTION_SIGNATURE_TYPE!);
+const SERVER = getServer(USER_FUNCTION!, SIGNATURE_TYPE!);
 const ERROR_HANDLER = new ErrorHandler(SERVER);
 SERVER.listen(PORT, () => {
   ERROR_HANDLER.register();
--- a/src/index.ts
+++ b/src/index.ts
@@ -44,7 +44,7 @@ const FLAG = {
 };
 
 // Supported environment variables
-const ENV_VARIABLE = {
+const ENV = {
   PORT: 'PORT',
   TARGET: 'TARGET',
   SIGNATURE_TYPE: 'SIGNATURE_TYPE', // underscore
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -28,30 +28,44 @@
...
+  PORT: 'port',
+  TARGET: 'target',
+  SIGNATURE_TYPE: 'signature-type', // dash
+};
+
+// Supported environment variables
+const ENV_VARIABLE = {
+  PORT: 'FUNCTION_PORT',
+  TARGET: 'FUNCTION_TARGET',
+  SIGNATURE_TYPE: 'FUNCTION_SIGNATURE_TYPE', // underscore
```
swalkowski: While we are refactoring this, please replace `FUNCTION_SIGNATURE_TYPE` in `FUNCTION_SIGNATURE_TYPE must be one of 'http' or 'event'.` message below with `Function signature type`, as it can refer to either flag or env var.

src/index.ts
```diff
@@ -28,30 +28,44 @@
...
 const argv = minimist(process.argv, {
-  string: ['port', 'function-target', 'function-signature-type'],
+  string: [ENV_VARIABLE.PORT, FLAG.TARGET, FLAG.SIGNATURE_TYPE],
 });
 
 const CODE_LOCATION = process.cwd();
-const PORT = argv['port'] || process.env.PORT || '8080';
+const PORT =
+  argv[ENV_VARIABLE.PORT] || process.env[ENV_VARIABLE.PORT] || '8080';
 const FUNCTION_TARGET =
```
swalkowski: While I can see how the refactoring you are proposing makes sense, it also makes the code a bit less readable, in my opinion, because of the extra mappings. For example, we are mapping `FUNCTION_TARGET` string to `ENV_VARIABLE.TARGET` and then resolving it to `FUNCTION_TARGET` const.

If the names without 'function' prefix are to be the canonical ones, then let's drop `FUNCTION_` prefix from `FUNCTION_TARGET`, `FUNCTION_SIGNATURE_TYPE_STRING`, `FUNCTION_SIGNATURE_TYPE` consts.

src/index.ts
```diff
@@ -28,30 +28,44 @@
...
+  SIGNATURE_TYPE: 'FUNCTION_SIGNATURE_TYPE', // underscore
+};
+
 enum NodeEnv {
   PRODUCTION = 'production',
 }
 
 const argv = minimist(process.argv, {
-  string: ['port', 'function-target', 'function-signature-type'],
+  string: [ENV_VARIABLE.PORT, FLAG.TARGET, FLAG.SIGNATURE_TYPE],
```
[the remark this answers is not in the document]
grant: Done.

src/index.ts
grant: Dropped the `FUNCTION_` prefix.

src/index.ts
```diff
@@ -28,30 +28,44 @@
...
+  argv[ENV_VARIABLE.PORT] || process.env[ENV_VARIABLE.PORT] || '8080';
 const FUNCTION_TARGET =
-  argv['function-target'] || process.env.FUNCTION_TARGET || 'function';
+  argv[FLAG.TARGET] || process.env[ENV_VARIABLE.TARGET] || 'function';
 
 const FUNCTION_SIGNATURE_TYPE_STRING =
-  argv['function-signature-type'] ||
-  process.env.FUNCTION_SIGNATURE_TYPE ||
-  'http';
+  argv[FLAG.SIGNATURE_TYPE] || process.env.FUNCTION_SIGNATURE_TYPE || 'http';
```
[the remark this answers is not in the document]
grant: Done.

src/index.ts
```diff
@@ -28,30 +28,44 @@
...
 
 const argv = minimist(process.argv, {
-  string: ['port', 'function-target', 'function-signature-type'],
+  string: [ENV_VARIABLE.PORT, FLAG.TARGET, FLAG.SIGNATURE_TYPE],
 });
 
 const CODE_LOCATION = process.cwd();
-const PORT = argv['port'] || process.env.PORT || '8080';
+const PORT =
+  argv[ENV_VARIABLE.PORT] || process.env[ENV_VARIABLE.PORT] || '8080';
```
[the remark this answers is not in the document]
grant: Fixed.

src/index.ts
```diff
@@ -28,40 +28,52 @@
...
 });
 
 const CODE_LOCATION = process.cwd();
-const PORT = argv['port'] || process.env.PORT || '8080';
-const FUNCTION_TARGET =
-  argv['function-target'] || process.env.FUNCTION_TARGET || 'function';
+const PORT = argv[FLAG.PORT] || process.env[ENV.PORT] || '8080';
+const TARGET = argv[FLAG.TARGET] || process.env[ENV.TARGET] || 'function';
 
 const FUNCTION_SIGNATURE_TYPE_STRING =
```
swalkowski: We could drop `FUNCTION_` prefix in this const as well, for consistency and brevity.

src/index.ts
```diff
@@ -28,40 +28,52 @@
...
-  argv['function-signature-type'] ||
-  process.env.FUNCTION_SIGNATURE_TYPE ||
-  'http';
+  argv[FLAG.SIGNATURE_TYPE] || process.env[ENV.SIGNATURE_TYPE] || 'http';
 const FUNCTION_SIGNATURE_TYPE =
```
swalkowski: We could drop `FUNCTION_` prefix in this const as well, for consistency and brevity.

src/index.ts
```diff
@@ -28,40 +28,52 @@
...
+const FLAG = {
+  PORT: 'port',
+  TARGET: 'target',
+  SIGNATURE_TYPE: 'signature-type', // dash
+};
+
+// Supported environment variables
+const ENV = {
+  PORT: 'PORT',
+  TARGET: 'TARGET',
```
swalkowski: Perhaps I wasn't clear in my previous comments: we should keep the environment variable names unchanged, just like in the table in README. Please keep PORT, FUNCTION_TARGET, FUNCTION_SIGNATURE_TYPE.

Short names are fine for flag names and const names in the code.

swalkowski: I mean:

```
const ENV = {
  PORT: 'PORT',
  TARGET: 'FUNCTION_TARGET',
  SIGNATURE_TYPE: 'FUNCTION_SIGNATURE_TYPE', // underscore
};
```

src/index.ts
grant: This request is getting out of scope, but I've changed it as I agree.

src/index.ts
grant: Done.

src/index.ts
grant: Yes. Agreed.
Not sure why I changed this. Reverted the `FUNCTION_` prefix for these env vars.

src/invoker.ts
```diff
@@ -23,8 +23,8 @@
 import * as bodyParser from 'body-parser';
 import * as domain from 'domain';
 import * as express from 'express';
-import * as onFinished from 'on-finished';
 import * as http from 'http';
+import * as onFinished from 'on-finished';
```
grant: IDE auto-sorted.

test/invoker.ts
```diff
@@ -13,10 +13,9 @@
 // limitations under the License.
 
 import * as assert from 'assert';
-import * as supertest from 'supertest';
 import * as express from 'express';
-
 import * as invoker from '../src/invoker';
+import * as supertest from 'supertest';
```
grant: IDE auto-sorted.

PR #18, opened 2019-04-18
docs(README): Replace --target with --function-target

anishkny:
Otherwise it errors:

```
$ npx @google-cloud/functions-framework --target=helloWorld
Function 'function' is not defined in the provided module.
Did you specify the correct target function to execute?
Could not load the function, shutting down.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ exports.helloWorld = (req, res) => {
 Run the following command:
 
 ```sh
-npx @google-cloud/functions-framework --target=helloWorld
+npx @google-cloud/functions-framework --function-target=helloWorld
 ```
 
 Open http://localhost:8080/ in your browser and see *Hello, World*.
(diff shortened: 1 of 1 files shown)


grant:
Please create an issue before a PR.

It looks like we need to publish to npm for the README instructions to be correct.

You're looking at the [previous version on npm](https://www.npmjs.com/package/@google-cloud/functions-framework). `master` in this repo has instructions.

---

We should publish the npm package after merging commits @swalkowski.

anishkny:
Filed issue. Fixes #19 

I was looking at README on master on GitHub [here](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/README.md).

stew-r:
We appreciate the issue & PR -- thanks for proactively finding a solution. Unfortunately, this is currently a mismatch between the (correct) docs and the latest version of our package. We're moving to 1.1.0 which replaces `function-target` with the shorter `target` CLI arg.

Based on that, I will close this PR.

anishkny:
Sure no hard feelings. Just trying to help out :wink:

stew-r:
👍

PR #20, opened 2019-04-18
Update index.ts comments

stew-r:
1. Remove extraneous env var docs
2. Clean up the intro comment ("user's code" -> "function code")

Note: if we want to keep the env var docs, we should probably document the CLI args as well.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,18 +14,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Functions framework entry point that configures and starts Node.js server
-// that runs user's code on HTTP request.
-// The following environment variables can be set to configure the framework:
-//   - PORT - defines the port on which this server listens to all HTTP
-//     requests.
-//   - FUNCTION_TARGET - defines the name of the function within user's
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -14,18 +14,8 @@
...
-//   - PORT - defines the port on which this server listens to all HTTP
-//     requests.
-//   - FUNCTION_TARGET - defines the name of the function within user's
-//     node module to execute. If such a function is not defined,
-//     then falls back to 'function' name.
-//   - FUNCTION_SIGNATURE_TYPE - defines the type of the client function
-//     signature, 'http' for function signature with HTTP request and HTTP
-//     response arguments, or 'event' for function signature with arguments
-//     unmarshalled from an incoming request.
+// Functions Framework entrypoint configures and starts a Node.js server
```
swalkowski: Nit: I would keep it as 'that configures' so that this comment clearly answers the question 'what is this?', and is in line with a similar comment in invoker.ts.

grant: @stew-r Do you want to follow up on this PR?
I think the working is fine in this PR, although I'd say "The Functions Framework ..."

PR #26, opened 2019-05-10
add a code location flag

anonrose:


grant:
Could we open an issue for this PR? I think there's more discussion needed before deciding on a flag.

For example, why is this location needed? Why can't we just run the FF in the code location?

stew-r:
+1 to @grant 

@anonrose what's the use case/what are you trying to solve with this flag?

grant:
Closing due to no PR description, and no issue report.
Please create an issue describing the use case first.

Issue #37, opened 2019-06-07
Documentation: Unclear how to use events signature-type and test pubsub


comolongo:
Hi,
I was able to setup and do local testing using http signature. However my goal is to do local development on a cloud function that consumes PubSub. What is the best development workflow for this?

I was thinking I may need to set signature-type to event, and somehow pass in a mock pubsub event object, how the current documentation doesn't give me many clues. Could someone point me the way?

Thanks!

grant:
Hey @comolongo, 
Can you see if this sample helps you?
https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/functions/pubsub/test/index.test.js#L32

Looks like it uses the HTTP signature, but it could be useful for a local dev workflow.
Maybe @ace-n could provide more detail.

Sabyasachi2k:
hi,
Any example for the signature-type event would really help. that example is for publishing. I have the subscribe function and I am unable to call the function to test it.

grant:
@Sabyasachi2k Does this guide help?

https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/40/files?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8

We're working on an event guide. I agree it would be useful.

Sabyasachi2k:
@grant Thank you for the link.
Go one question. 
1. I am able to call the function using the instruction given, however instead of the variable being available in the pubSubEvent variable, its available on the context variable. Any idea what needs to be done
function helloPubSub(pubSubEvent, _context) {
console.log (" the data passed is " + JSON.stringify(pubSubEvent) + ", " + JSON.stringify(_context));
}

grant:
> I am able to call the function using the instruction given, however instead of the variable being available in the pubSubEvent variable, its available on the context variable. Any idea what needs to be done

I don't know at the moment.

Node functions for Pub/Sub triggers should look like this:
https://cloud.google.com/functions/docs/calling/pubsub#sample_code

If we think there is a bug with following that code, we should file a bug with expected/actual behavior that might be in this framework or elsewhere.

Sabyasachi2k:
Thanks a lot @grant 
Is there a location where the bug report can be done.

grant:
> Thanks a lot @grant
> Is there a location where the bug report can be done.

There's a **feedback** button on the page for creating the bug report.

PR #40, opened 2019-06-17
Added new section in README.md to clarify how to do local testing of …

comolongo:
Added new section in README.md to clarify how to do local testing of event functions. Fixes the questions raised in issue #37.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ documentation on
 [background functions](https://cloud.google.com/functions/docs/writing/background#cloud_pubsub_example).
 
 ## Local testing of cloud events
-The setup for cloud functions that accept events is very similar to the instructions in the quickstart, with the following adjustments differences.
+The setup for cloud functions that accept events is very similar to the instructions in the quickstart, with the following adjustments:
 
 In your package.json, add a signature type (in bold) to your start command:
 <pre>
--- a/README.md
+++ b/README.md
@@ -179,6 +179,73 @@ For more details on this signature type, check out the Google Cloud Functions
 documentation on
 [background functions](https://cloud.google.com/functions/docs/writing/background#cloud_pubsub_example).
 
+## Local testing of cloud events
+The setup for cloud functions that accept events is very similar to the instructions in the quickstart, with the following adjustments differences.
+
+In your package.json, add a signature type (in bold) to your start command:
+<pre>
+  "scripts": {
+    "start": "functions-framework --target=helloWorld <b>--signature-type=event"</b>
+  }
+</pre>
+
+Upon running ```sh npm start ```, you'll see the function is still being served at http://localhost:8080/. However it is no longer accessible via GET requests from the browser. Instead, send a POST request where the request body conforms to the API defined by [push subscriptions](https://cloud.google.com/pubsub/docs/push). 
+
+### Submitting POST request to simulating a pubsub message
+
+Create mockPubsub.json file with the following contents:
+```json
+{
+  "message": {
+    "attributes": {
+      "key": "value"
+    },
+    "data": "SGVsbG8gQ2xvdWQgUHViL1N1YiEgSGVyZSBpcyBteSBtZXNzYWdlIQ==",
+    "messageId": "136969346945"
+  },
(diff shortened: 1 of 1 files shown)


stew-r:
This is a great contribution -- thank you!

@ace-n @grant think we can have this exist as a separate testing doc?

@comolongo have you tested whether this actually works with the Pub/Sub emulator? I think we have an issue right now where Pub/Sub messages are transformed in our infrastructure before reaching the container with your function. This means that the framework expects a slightly different message format.

To be clear: this is a bug that we need to fix. We just haven't gotten around to it yet.

grant:
@stew-r I recommend that we have a `docs/` folder in this repo for advanced documentation.

I can create that:
https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/42
Advanced guides should talk about things like using the emulator, multiple functions, etc.

As we gain more options for using the Functions Framework, this folder can help support docs.

comolongo:
Hey @stew-r , yes I have tested this with the PubSub emulator, and it does have the issues you're talking about. I was able to overcome it, by hacking the wrapEventFunction() in invoker.ts and filed ticket #41 to address that specifically.

On the topic of the PubSub emulator incorrectly transforming the messages, at least from my tests, it seems like the emulator is pushing notifications using the same format as on prod - i.e. it seems like the actual body of their push notification is correct. The issue may be that the emulator needs to include the right header flags for invoker::isBinaryCloudEvent() to return true. Currently it is returning false, and causing wrapEventFunction() to go down the wrong else clause.

stew-r:
@grant I like the structure -- thanks!

@comolongo so I think the issue is somewhat different (though your solution addresses it too). In short:
* Pub/Sub sends [a message](https://cloud.google.com/pubsub/docs/push#receiving_push_messages)
* In Google Cloud Functions, we have a piece of infrastructure that receives the message and applies some transformation before passing it on to the user's container
* The user's container thus doesn't see a Pub/Sub message but a transformed Pub/Sub message

When running locally (or a different container-based environment), we don't perform this transformation. So your function sees something different based on whether it is running in Google Cloud Functions vs. elsewhere.

As you've discovered, this is solvable. Thanks for submitting a fix, I'll make sure to pass it along to the eng team.

grant:
@comolongo Could you change this PR to add this content into the `docs` folder?

dsnowdon:
Might be handy for typescript users if it's stated that the entry file is dictated by your `  "main": "./lib/index.js",` property in `package.json`?

It just seemed a little odd to me that I couldn't add a argument to the command, most testing CLIs follow that flow.

grant:
@comolongo Could you move markdown content into the `docs/` folder?

I'm trying to resolve PRs within 30 days. If there's no reply, we should add the guides outside of this PR.

README.md
```diff
@@ -179,6 +179,73 @@ For more details on this signature type, check out the Google Cloud Functions
...
+    "data": "SGVsbG8gQ2xvdWQgUHViL1N1YiEgSGVyZSBpcyBteSBtZXNzYWdlIQ==",
+    "messageId": "136969346945"
+  },
+  "subscription": "projects/myproject/subscriptions/mysubscription"
+}
+```
+
+The file can be in any folder on your computer. From the terminal, goto the directory where ```mockPubsub.json``` is located, and run the following command assuming your cloud function is hosted locally on port 8080:
+```sh 
+curl -d "@mockPubsub.json" -X POST http://localhost:8080
```
betacar: May I also suggest to add the `CE-*` headers, maybe with a _truthy_ value, to the docs so the lib [can know](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/src/invoker.ts#L379-L386) that the request is for a Cloud Event?

```sh
curl -d "@mockPubsub.json" -X POST -H "Ce-Type: true" -H "Ce-Specversion: true" -H "Ce-Source: true" -H "Ce-Id: true" http://localhost:8080
```

This would help with issues like #41, and also set the values correctly to the `data` and `context` arguments.

grant:
Hey @comolongo,
Can you move the change here to the `docs/` folder? I'll merge this right after.

grant:
Closed due to inactivity.

PR #55, opened 2019-07-18
Add support for installing dependencies through yarn

lukasberger:
Function dependencies are now installed using yarn when yarn.lock is
present in the application source code.


stew-r:
Nice!

Issue #57, opened 2019-07-18
Stackdriver Trace Integration


grant:
## Stackdriver Trace Integration

Add Stackdriver Trace support to the Functions Framework.

See: https://github.com/googleapis/cloud-trace-nodejs

> @stew-r Can you add details as to what this would do? Would this be a tutorial or part of the actual framework?

stew-r:
Ideally, this would be part of the Functions Framework and a user of the framework gets trace support built in automatically. The user does not need to think about importing a client or doing any setup. They can disable this feature by passing some configuration.

grant:
I added a WIP PR for this feature.

- Should these tools be added as an explicit dependency to the FF? (Is that an acceptable increase in GCF dependencies)?
- Should the disable flag be something like:
  - `--disable-stackdriver`, `FUNCTION_DISABLE_STACKDRIVER`

Elyx0:
Struggling aswell here from https://github.com/googleapis/nodejs-logging/issues/591 any updates?

grant:
So far there has been no priority for this feature, no upvotes.
The issue is that `require('@google-cloud/trace-agent').start();` must be the first module loaded.

I don't think it would be hard to implement that, but we need someone to help implement a good PR.

Elyx0:
Can I add this myself @grant, does it need to be the first module in the index.src exporting the cloud functions or required inside every functions?

dinigo:
As I understad it, this feature was available at some point, but was removed afterwards, and now it's being re-implemented in the FF?

What kind of help do you need in the PR? I may be able to savage some time from here and there. We (Divisadero) rely on Cloud Functions and want more visibility of what's happening within executions

grant:
I'm not an expert in Stackdriver Trace so I'm not sure of the requirements for this.

We need a proposed implementation for optional trace support that would not affect the performance of all functions that don't enable it. Then we need a PR that follows that proposal that we can get approved by the eng team.

I think GCF might already have trace support, I'm not sure.

manwithsteelnerves:
Any update on this?
Can someone suggest how can I assure **require('@google-cloud/trace-agent').start();** will be loaded at start?

manwithsteelnerves:
Can someone give a hint on how to consider the stack trace logging in the above image?

grant:
I haven't looked at this since.
I would assume the steps would be:
- Ensure a hello world service on Cloud Run works
[...]

PR #59, opened 2019-07-19
WIP: feat: Add Stackdriver integration

grant:
WIP: Adds StackDriver Debug, Trace, and Profiler support.

## TODO:
- Add an env var option for disabling the integration
- Fix the tests
  - Add setup to Travis: https://www.npmjs.com/package/@google-cloud/profiler#basic-set-up
    - Create GOOGLE_APPLICATION_CREDENTIALS
  - Error: @google-cloud/profiler Failed to create profile
    -  request to HTTP function
      - ```
        @google-cloud/profiler Failed to create profile, waiting 9.3s to try again: Error: Could not refresh access token: A Not Found error was returned while attempting to retrieve an accesstoken for the Compute Engine built-in service account. This may be because the Compute Engine instance does not have any permission scopes specified: Could not refresh access token: Unsuccessful response status code. Request failed with status code 404
        @google-cloud/debug-agent Failed to re-register debuggee eco-emissary-99515: Error: Could not refresh access token: A Not Found error was returned while attempting to retrieve an accesstoken for the Compute Engine built-in service account. This may be because the Compute Engine instance does not have any permission scopes specified: Could not refresh access token: Unsuccessful response status code. Request failed with status code 404
        1) should return transformed body
        ```

## Open Questions
- Should these be normal dependancies rather than devDependancies?
- How can we test this PR?

**Fixes:** #57

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,9 @@
   "author": "Google Inc.",
   "license": "Apache-2.0",
   "devDependencies": {
+    "@google-cloud/debug-agent": "^4.0.1",
+    "@google-cloud/profiler": "^2.0.2",
+    "@google-cloud/trace-agent": "^4.1.0",
     "@types/body-parser": "^1.17.0",
     "@types/express": "^4.16.1",
     "@types/minimist": "^1.2.0",
--- a/src/invoker.ts
+++ b/src/invoker.ts
@@ -26,6 +26,8 @@ import * as express from 'express';
 import * as http from 'http';
 import * as onFinished from 'on-finished';
 
+import { startTooling } from './tools';
+
 // HTTP header field that is added to Worker response to signalize problems with
 // executing the client function.
 const FUNCTION_STATUS_HEADER_FIELD = 'X-Google-Status';
@@ -358,7 +360,10 @@ function makeHttpHandler(execute: HttpFunction): express.RequestHandler {
         logAndSendError(err, res);
       }
     });
-    d.run(() => {
+    d.run(async () => {
+      // Start the debugger (if not alreay started)
+      const tools = startTooling();
+      if (tools) await tools.debug.isReady();
       process.nextTick(() => {
         execute(req, res);
       });
--- a/src/tools.ts
+++ b/src/tools.ts
@@ -1,0 +1,60 @@
+import * as debugAgent from '@google-cloud/debug-agent';
+import * as profiler from '@google-cloud/profiler';
+import * as trace from '@google-cloud/trace-agent';
+
+let toolsAreSetup = false;
+
+/**
+ * Starts Google Cloud tooling.
+ * - StackDriver Trace
+ * - StackDriver Profiler
+ * - StackDriver Debugger
+ */
+export function startTooling() {
+  // Ensure tools are only setup once per process.
+  if (toolsAreSetup) {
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -34,6 +34,9 @@
   "author": "Google Inc.",
   "license": "Apache-2.0",
   "devDependencies": {
```
squee1945: devDependencies will not end up in the build application.

src/tools.ts
```diff
@@ -0,0 +1,60 @@
...
+export function startTooling() {
+  // Ensure tools are only setup once per process.
+  if (toolsAreSetup) {
+    return null;
+  } else {
+    toolsAreSetup = true;
+  }
+
+  // Start Trace
+  trace.start();
```
squee1945: I thought the trace-agent needed to be started as the very first part of the app:
```
require('@google-cloud/trace-agent').start();
```
See https://www.npmjs.com/package/@google-cloud/trace-agent#usage

src/invoker.ts
```diff
@@ -358,7 +360,10 @@ function makeHttpHandler(execute: HttpFunction): express.RequestHandler {
         logAndSendError(err, res);
       }
     });
-    d.run(() => {
+    d.run(async () => {
+      // Start the debugger (if not alreay started)
+      const tools = startTooling();
```
squee1945: There is overhead both in starting the tooling, and ongoing performance overhead. Is there a way for the user to opt out of this?

src/invoker.ts
```diff
@@ -26,6 +26,8 @@ import * as express from 'express';
 import * as http from 'http';
 import * as onFinished from 'on-finished';
 
+import { startTooling } from './tools';
```
squee1945: Nit: `diagnostics` instead of `tools`?

src/invoker.ts
grayside: For ease in turning instrumentation on and off, it would be nice to toggle this with an environment variable instead of something requiring a redeployment (e.g., adjusted container entrypoint)

stew-r: Yes. The PRD specifies the following semantics:

**Enable/disable Stackdriver APM**
As a developer, I can enable or disable individual Stackdriver APM features. The on/off state is controlled by an environment variable as follows:

Stackdriver APM feature | Environment variable
-- | --
Stackdriver Trace | GOOGLE_CLOUD_TRACE_ENABLED

If the environment variable is set to `false` or if it is not set, the feature should be disabled by default.

**By default, Stackdriver APM features are enabled on Cloud Functions**
As a developer on Cloud Functions, when I deploy my function Stackdriver Trace is enabled by default. The container into which my function is deployed will, e.g., automatically set the value of the `GOOGLE_CLOUD_TRACE_ENABLED` environment variable to `true`.

src/tools.ts
```diff
@@ -0,0 +1,60 @@
+import * as debugAgent from '@google-cloud/debug-agent';
+import * as profiler from '@google-cloud/profiler';
```
grayside: I understand that profiler currently has limitations on working with musl.

grant:
This PR needs more of a design doc and solution across all FFs. Let's continue in the GitHub issue.

PR #72, opened 2019-08-07
Remove the converter

hdp617:
The converter should not be part of the functions framework. It is not necessary for local development or deployment.


squee1945:
Let's not merge this P-R yet. Is there any reason to do this at this point?

hdp617:
My rationale is to have a single source of truth for the converter.

PR #81, opened 2019-09-02
fixes #33: Only listen to functionTarget.

### LINKED ISSUE #33: functions called from browser execute twice
Although curl will only call the function once.
Chrome load or refresh calls the function twice.
A bit confusing when debugging something more complex than HelloWorld.

Very easy to see:
```
exports.helloWorld = (req, res) => {
    console.log('running HelloWorld'); // <-- oh no not again
    res.send('Hello, World');
  };
```

Is there a way to stop this?

grant:
- Specifies _exact_ function target instead of `*`.
- Fixes #33.
- Ran linter.

### DIFF
--- a/test/invoker.ts
+++ b/test/invoker.ts
@@ -35,7 +35,7 @@ describe('request to HTTP function', () => {
         res.send(req.body.text.toUpperCase());
       },
       invoker.SignatureType.HTTP,
-      'TARGET'
+      '/'
     );
     return supertest(server)
       .post('/')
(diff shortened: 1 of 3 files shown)


hdp617:
Could you also update the test?

grant:
> Could you also update the test?
@hdp617 Addressed comment and fixed Travis CI tests.

PTAL.

test/invoker.ts
```diff
@@ -34,7 +34,8 @@ describe('request to HTTP function', () => {
       (req: express.Request, res: express.Response) => {
         res.send(req.body.text.toUpperCase());
       },
-      invoker.SignatureType.HTTP
+      invoker.SignatureType.HTTP,
+      '/'
```
hdp617: This is an odd function target. I suggest something like 'testFunction'.

test/invoker.ts
```diff
@@ -34,7 +34,8 @@ describe('request to HTTP function', () => {
...
     );
     return supertest(server)
       .post('/')
```
hdp617: `.post('/'testFunction')`?

test/invoker.ts
grant: Applied suggestion of `testFunction`.

test/invoker.ts
grant: Done.

Issue #89, opened 2019-09-13
The "path" argument must be of type string. Received type undefined @1.3.0


jpbarbosa:
```
npx @google-cloud/functions-framework --target=helloWorld
The "path" argument must be of type string. Received type undefined
```

I notice the problem at version 1.3.0. At version 1.2.1 the command runs normally.

My system is MacOS 10.14.6, with Node 10.16.3 installed with Homebrew.

dfanchon:
Having the same issue here with a cloud build deployment.

timswalling:
When I add `--source=./` to the arguments it removes the error message but I then get a 404 (`Cannot GET /`) when trying to trigger the function

yvmarques:
I got the same error, and I fixed with the following flag at the start script.

`functions-framework --target=functionName --source='./'`

I guess @timswalling  you need to enclose the `./` with quotes in order to have a string

hdp617:
I'm sorry for this bad release. I'll make a new release soon. In the mean time, you can remove functions-framework as a dependency in your `package.json` when you deploy to GCF. I think it will fix this issue.

PR #90, opened 2019-09-13
fix: use empty string path when function source is not specified

hdp617:
I removed `'/'` from `CODE_LOCATION` that set the default function directory to root in commit #77. This causes this error in 1.3.0,
```
npx @google-cloud/functions-framework --target=helloWorld
The "path" argument must be of type string. Received type undefined
```

When `--source` flag and `FUNCTION_SOURCE` env var are not specified, one of the 'path' arguments in `resolve()` is undefined. This causes the error above.

I've added empty string as an argument in the case above.

fixes issue #89

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -63,7 +63,7 @@ const argv = minimist(process.argv, {
 
 const CODE_LOCATION = resolve(
   process.cwd(),
-  argv[FLAG.SOURCE] || process.env[ENV.SOURCE]
+  argv[FLAG.SOURCE] || process.env[ENV.SOURCE] || ''
 );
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -63,7 +63,7 @@ const argv = minimist(process.argv, {
 
 const CODE_LOCATION = resolve(
   process.cwd(),
```
lukasberger: Isn't process.cwd redundant here? According to the docs [1] the cwd is prepended to any relative path.

https://nodejs.org/docs/latest/api/path.html#path_path_resolve_paths

hdp617: It is redundant. I thought the docs meant `cwd` would be used as a path instead of `cwd` being prepended.

PR #106, opened 2019-11-13
Add cookieParser middleware

Tebro:
You can already set cookies, but they are not parsed into the Request without this middleware.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@google-cloud/functions-framework",
-  "version": "1.2.1",
+  "version": "1.3.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


grant:
Huy, what do you think?
This cookie parsing middleware is quite common in Express applications.
It's 1 dependency, 94kb, shouldn't be a huge impact on GCF.

squee1945:
What is the use case for a function to read cookies?

Tebro:
My current use case is for a OICD flow for my Slack application. (To couple slack users to O365 users).

Need to persist some user information in some kind of session and cookies was a convenient way to do that.

squee1945:
The HTTP headers are available. Can you use a cookie package against the header directly? I.e., why does it need to be middleware?

grant:
I'd guess you'd be able add the Express middleware like you can do with the Express Routing middleware:

I have a blogpost on how to add middleware. Here are some useful links:
- https://medium.com/google-cloud/express-routing-with-google-cloud-functions-36fb55885c68
- https://github.com/expressjs/cookie-parser#api

Tebro:
Hmm, that would probably work as well if the functions framework wants leave this out.

grant:
Yeah, the FF is not including anything that is not essential, including Express' cookie parsing middleware. The FF really should only include essential features that can be implemented in all languages very easily. For example, if some GCF developers don't want this feature, we shouldn't force them to use it and increase their module size.

grant:
I'll look into writing a blogpost about Cookie parsing with the FF.

grant:
We're going to let users parse cookies from the `request` object, and not add this Express middleware.

PR #107, opened 2019-11-18
Fix double quotation from full-width character

jupemara:
So far, Just only convert from full-width double quotation (`”`) to single-width double quotation (`"`) .


hdp617:
Thanks!

jupemara:
Thanks for merging!

PR #109, opened 2019-11-22
docs: update pub sub docs with instructions that work

### LINKED ISSUE #85: Can't get local simulated events unmarshalled correctly after following the latest docs
Hi everyone,

I've been trying to mock a pub/sub event on my local dev according to the docs at https://github.com/GoogleCloudPlatform/functions-framework-nodejs/tree/master/docs#local-testing-of-cloud-events. Nothing special, I'm just trying to manipulate the `data` object passed as a parameter to the function (which I assume should be unmarshalled, given that I'm starting the server with `--signature-type=event`).

Function code:
`exports.helloPubSub = (data, context) => console.log(data);`

Mock Json:
```
{
  "message": {
    "attributes": {
      "key": "value"
    },
    "data": "SGVsbG8gQ2xvdWQgUHViL1N1YiEgSGVyZSBpcyBteSBtZXNzYWdlIQ==",
    "messageId": "136969346945"
  },
  "subscription": "projects/myproject/subscriptions/mysubscription"
}
```

Output
```
{
  '{  "message": {    "attributes": {      "key": "value"    },    "data": "SGVsbG8gQ2xvdWQgUHViL1N1YiEgSGVyZSBpcyBteSBtZXNzYWdlIQ': '=",    "messageId": "136969346945"  },  "subscription": "projects/myproject/subscriptions/mysubscription"}'
```

All of the above is pretty standard. However, I can't get to parse the resulting object in any meaningful way.

Would you be able to shed some light on this issue?

grant:
Fixes user submitted documentation about using the Pub/Sub emulator.
Tested locally.
Fixes #85
R: @grayside

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -1,20 +1,44 @@
-## Local Testing of Cloud Events
+# Events
 
-The setup for cloud functions that accept events is very similar to the instructions in the quickstart, with the following adjustments:
+This guide shows you how to use the Functions Framework for local testing with:
 
-In your `package.json`, add a signature type (in bold) to your start command:
+- CloudEvents
+- The Pub/Sub Emulator
 
-<pre>
+## Local Testing of CloudEvents
+
+In your `package.json`, specify `--signature-type=event"` for the `functions-framework`:
+
+```sh
+{
   "scripts": {
-    "start": "functions-framework --target=helloWorld <b>--signature-type=event"</b>
+    "start": "functions-framework --target=helloWorld --signature-type=event"
   }
-</pre>
+}
+```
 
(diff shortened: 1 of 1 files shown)


docs/events.md
```diff
@@ -1,20 +1,44 @@
...
+
+Start the Functions Framework:
+
+```sh
+npm start
+```
 
-Create a `mockPubsub.json` file with the following contents:
+Your function will be serving at `http://localhost:8080/`, however,
+it is no longer accessible via `HTTP GET` requests from the browser.
```
grayside: What happens if you send it a GET request?

grant: You get a standard HTTP 404 response with Express' default `Cannot GET ${PATH}` error. I don't think this is worth additionally documenting/reading.

PR #117, opened 2020-01-27
feat: add cloudevents module

grant:
Use a Node CloudEvents SDK instead of inlining the CE interface within this package.

Does not affect runtime, just types.


grant:
Let's not add this now as the module is still in development.

PR #118, opened 2020-01-28
feat: add --dry-run option to load user's function but not start a server

hdp617:
With this option, we can load a function without starting a server, i.e.
```
functions-framework --target=helloWorld --dry-run=true
```
so that it can be used in builds to check for syntax errors, etc.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -43,6 +43,7 @@ const FLAG = {
   TARGET: 'target',
   SIGNATURE_TYPE: 'signature-type', // dash
   SOURCE: 'source',
+  DRY_RUN: 'dry-run',
 };
 
 // Supported environment variables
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -97,6 +100,14 @@ if (!USER_FUNCTION) {
 
 const SERVER = getServer(USER_FUNCTION!, SIGNATURE_TYPE!);
 const ERROR_HANDLER = new ErrorHandler(SERVER);
+
+if (DRY_RUN) {
+  console.log(`Function: ${TARGET}`);
+  console.log(`URL: http://localhost:${PORT}/`);
+  console.log('Dry run successfully, shutting down.');
```
lukasberger: nit: successful

src/index.ts
```diff
@@ -97,6 +100,14 @@ if (!USER_FUNCTION) {
...
+  process.exit(0);
```
lukasberger: You could potentially move the process.exit(0) inside the SERVER.listen and take advantage of the console.logs there, but either is fine.

PR #121, opened 2020-03-03
Adjust path handling

hdp617:
- paths with favicon.ico and robots.txt will send 404 and no longer invoke function.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -43,7 +43,6 @@ const FLAG = {
   TARGET: 'target',
   SIGNATURE_TYPE: 'signature-type', // dash
   SOURCE: 'source',
-  DRY_RUN: 'dry-run',
 };
 
 // Supported environment variables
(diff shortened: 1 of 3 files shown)


TexasFreedom:
please create a friendly way for me to participate in whatever necessary security requests you have, i have done all in good faith with clean hands and if a mistake or misunderstanding or inconvenience exists please regard substance over form as i am a swim coach and only interacted with ai system platform to help design a system which would verify truth through dashboard audit and ledger as a witness. 

I have lost most everything due to corruption and although i am a victim of severe abuse, my dedication is to intentional solutions which protect the vulnerable from exploitive abusive prctices. 

Equity shall not suffer a wrong to be without a remedy. Out of pure despartion to provide truth through better solutions, Requested ai to design a system dashboard and a distribution system in access to protective services and support. i am unfortuanately being completely accused of everything ill willed, but i the facts remain a severe injustice continues without any remedy. 

what should i do? 
Andie SHA256

Issue #96, opened 2019-09-17
Event parsing doesn't match Cloud Functions in background mode


ericnorris:
Hey there!

Related to #41 and #85, I ran into a lot of confusion trying to sort out what the format should be for the request body to a local [background function](https://cloud.google.com/functions/docs/writing/background).

In production a background function has the following for the first parameter:

```
{ @type: 'type.googleapis.com/google.pubsub.v1.PubsubMessage',
  attributes: null,
  data:  '<base64 data>' }
```

and the following for the "context" parameter:

```
{ eventId: '749799261766789',
  timestamp: '2019-09-17T19:56:09.018Z',
  eventType: 'providers/cloud.pubsub/eventTypes/topic.publish',
  resource: '<pubsub topic>' }
```

The [current docs](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/de6ef3874ae703084d5b7566ecde743a8f80bd36/docs/events.md) for using the functions framework for events suggests `curl`ing data in a format that results in the parameters not matching the above - the first parameter becomes:

```
{ message:
   { attributes: {},
     data: '<base64 data>',
     messageId: '136969346945' },
  subscription: 'projects/myproject/subscriptions/mysubscription' }
```

And the "context" parameter:

```
{ type: 'true',
  specversion: 'true',
  source: 'true',
  id: 'true' }
```

Digging into the source code, https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/de6ef3874ae703084d5b7566ecde743a8f80bd36/src/invoker.ts#L431-L447 seems to expect the body (for "legacy events") to look something like:

```
{ data:
  { data: '<base64 data>`,
    attributes: null,
    @type: 'type.googleapis.com/google.pubsub.v1.PubsubMessage' },
  eventId: 749799261766789 }
```

(Further evidence this is expected is in this [WIP branch](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/1255/files#diff-a5c7e60c1ae9d8d39557244c08a97e48R47) for NodeJS samples.)

Using the above POST body seems to result in function parameters that most closely match production, and although it doesn't match [the POST body for push messages](https://cloud.google.com/pubsub/docs/push#receiving_push_messages) I'm assuming that's what users should send if they want their function to act closest to what they will see when they deploy to Cloud Functions.

Is this correct?

grant:
[...]

Issue #41, opened 2019-06-17
Cloud events incorrectly unmarshalled: first argument is undefined, 2nd argument differs from production


comolongo:
When Pubsub notifications are sent to a locally hosted cloud event function, the first argument is undefined. It should contain the event information. The 2nd argument though present, does not match production signature; rather than containing metadata about the notification, it contains the entire event object, which should be the contents of argument 1 instead.

## My dummy cloud events function:

```js
exports.helloEvents = (data, context) => {
  console.log(data);
  console.log(context);
};
```

Pub/sub notification being sent:

```js
{
  "subscription": "projects/myproj/subscriptions/mysub",
  "message": {
    "data": "...my base 64 encoded data...",
    "messageId": "1",
    "attributes": {}
  }
}
```

## helloEvents log in production:
Argument 1, data, is:

```js
{  
   "subscription":"projects/myproj/subscriptions/mysub",
   "message":{  
      "data":"...my base 64 encoded data...",
      "messageId":"1",
      "attributes":{}
   }
}
```

Argument 2, context, is:
```js
{  
   "eventId":"12032d95-089a-4663-9be5-8aafd9d4c623",
   "resource":{  
      "service":"pubsub.googleapis.com",
      "name":"projects/myproj/topics/mytopic"
   },
   "eventType":"google.pubsub.topic.publish",
   "timestamp":"2019-06-16T18:39:44.809Z"
}
```

## On local, the results I get is very different.

- Argument 1: undefined
- Argument 2: 
```js
{  
   "subscription":"projects/myproj/subscriptions/mysub",
   "message":{  
      "data":"...my base 64 encoded data...",
      "messageId":"1",
      "attributes":{}
   }
}
```

comolongo:
From looking through the code, my problem is fixed if isBinaryCloudEvent() in wrapEventFunction always returned true at [line 433](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/src/invoker.ts#L433).

Currently the push notification from the pubsub emulator is causing the code to enter the else clause I that supports legacy events and CloudEvents in structured content mode. I'm not sure if this is an issue with the pubsub emulator, or if the support for legacy events can be removed, otherwise I'd be happy to push a CL for review.

burtonjc:
[...]

PR #125, opened 2020-03-19
Added GCP local parity

neatnerd:
There is a well-known bug with this framework documented in #96 and #41. The basic problem is that Cloud PubSub (at least local version) does not actually respect [Cloud Event Specs](https://github.com/cloudevents/spec/blob/v1.0/spec.md#message) in respect to binary vs structured mode by providing no respective HTTP (CE-) headers. 

The purpose of this framework is to enable local testing of GCP Cloud function, which is not currently not fulfilled for background function use case. Even example provided in docs would not currently work

I tried to carefully implement it and only do changes to the current logic only when it is absolutely necessary. I have also added a test to show a use case. For field specifications, I followed GCP documentation on [background functions](https://cloud.google.com/functions/docs/writing/background#function_parameters.)

### DIFF
--- a/src/invoker.ts
+++ b/src/invoker.ts
@@ -445,6 +445,21 @@ function wrapEventFunction(
       context.data = undefined;
       delete context.data;
     }
+    // Populating data for behavior compliant with parameter settings of Cloud PubSub
+    // https://cloud.google.com/functions/docs/writing/background#function_parameters
+    if (!data && event.message) {
+      data = event.message!.data;
+      if (!context.eventId) {
+        context.eventId = event.message.messageId;
+      }
+      if (!context.timestamp) {
+        context.timestamp =
+          event.message.publishTime || new Date().toISOString();
+      }
+      if (!context.eventType) {
+        context.eventType = 'google.pubsub.topic.publish';
+      }
+    }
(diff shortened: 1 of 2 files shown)


neatnerd:
@googlebot I signed it!

grant:
Thanks for the PR @neatnerd.
We do know that this is an issue and are actively planning a solution across the Function Frameworks.

Once we have the plan ready, I'll create an issue and hopefully we can use this PR or some other PRs to fix it.

anguillanneuf:
@grant Is this issue fixable through functions-framework? Or do you need Pub/Sub emulator to make it possible?

grant:
@anguillanneuf What's the specific issue you're referring to?
Can you comment on the GitHub issue? Or create a new GitHub issue if needed?

anguillanneuf:
@grant I emailed you about it. [Here](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/41#issuecomment-544054901) is the good description of it.

grant:
@anguillanneuf OK. I'll check my email.

We should continue conversation in the issue rather than this closed PR which I don't think will solve the issue. Unless you're suggesting we re-open this PR.

PR #126, opened 2020-03-25
fix: handle SIGINT in ErrorHandler

hdp617:
Currently, when running the Functions Framework in a Docker container by running `docker run`, one cannot stop the running container with Ctrl+C. This is because the Functions Framework doesn't handle SIGINT in the ErrorHandler. This change handles SIGINT in the similar fashion as SIGTERM i.e. close the web server and exit the process.

### DIFF
--- a/src/invoker.ts
+++ b/src/invoker.ts
@@ -535,9 +535,12 @@ export class ErrorHandler {
       logAndSendError(new Error(`Process exited with code ${code}`), latestRes);
     });
 
-    process.on('SIGTERM', () => {
-      this.server.close(() => {
-        process.exit();
+    ['SIGINT', 'SIGTERM'].forEach(signal => {
+      process.on(signal as NodeJS.Signals, () => {
+        console.log(`Received ${signal}`);
+        this.server.close(() => {
+          process.exit();
+        });
       });
     });
   }


l0co:
With this fix you can't now stop the server from the console (in Linux). After clicking CTRL+C it only says "Received SIGINT" forever. Exiting with CTRL+C works only after downgrading to 1.5.0.

hdp617:
I couldn't reproduce this on my Linux machine. How do you run the Functions Framework?

l0co:
Using `npx @google-cloud/functions-framework --target=webhookPublisher`

I've checked once again and it seems it's something with this code:

```js
        ['SIGINT', 'SIGTERM'].forEach(signal => {
            process.on(signal, () => {
                console.log(`Received ${signal}`);
                this.server.close(() => {
                    process.exit();
                });
            });
        });
```

To reproduce this problem after starting the server I need to make some requests to the function first, for example 10 quick requests one after another. I respond to them just with:

```
export async function webhookPublisher(req: Request, res: Response) {
        res
            .status(400)
            .type("text/plain")
            .send("Some error message")
            .end()
}
```

Here is the log when I try to kill it with `CTRL+C`:

```
$ npx @google-cloud/functions-framework --target=webhookPublisher
Serving function...
Function: webhookPublisher
URL: http://localhost:8080/
^CReceived SIGINT
^CReceived SIGINT
^CReceived SIGINT
^CReceived SIGINT
^CReceived SIGINT
^CReceived SIGINT
^CReceived SIGINT
^CReceived SIGINT
Error: Process exited with code 0
    at process.on.code (./node_modules/@google-cloud/functions-framework/build/src/invoker.js:396:29)
    at process.emit (events.js:189:13)
    at process.EventEmitter.emit (domain.js:441:20)
    at process.exit (internal/process/per_thread.js:168:15)
    at Server.server.close (./node_modules/@google-cloud/functions-framework/build/src/invoker.js:402:29)
    at Object.onceWrapper (events.js:277:13)
    at Server.emit (events.js:194:15)
    at Server.EventEmitter.emit (domain.js:441:20)
    at emitCloseNT (net.js:1609:8)
    at process._tickCallback (internal/process/next_tick.js:63:19)
```

If I don't do any request it exist immediately.

In `1.5.0` it exits immediately, even after requests.

But now I think there's the same problem in both `1.5.1` and `1.5.0` because the only difference is that `1.5.0` doesn't wait for anything and dies, while `1.5.1` waits for server stop. The problem seems to be with `server.stop()`.

Issue #134, opened 2020-05-06
Refactor invoker.ts functionality into separate files


grant:
The `invoker.ts` file of ~600 lines contains dozens of pieces of logic that are somewhat unrelated:
- Error handling
- Serving
- Routing
- Eventing
- Logging

This makes the framework difficult to understand and introduces many "global" variables. It would be much easier to separate some of these functions out into separate files.

hdp617:
Thanks for bringing this up Grant! I think all the logic to load user's function like `getUserFunction` can be moved to its own file as well.

PR #136, opened 2020-05-07
refactor: Move the logic to load user's function to loader.ts

hdp617:
This separates the logic to load user's function and one to start the webserver. Moves the function signature type definition to function.ts. Also, adds more testcases for the function loading logic.

See also: #134


test/data/with_main/foo.js
```diff
@@ -0,0 +1,24 @@
+/**
+ *
```
grant, line 2: These comments are a bit weird, even if they are in test files.

hdp617: Oops. Added comments.

test/loader.ts
```diff
@@ -0,0 +1,50 @@
...
+      target: 'testEventFunction',
+    },
+    {
+      name: 'function with "main" in package.json',
+      codeLocation: '/test/data/with_main',
+      target: 'testEventFunction',
+    },
+  ];
+
+  testData.forEach(test => {
```
grant, line 40: `for (const test of testData) {` removes the unnecessary closure.

PR #138, opened 2020-05-12
refactor: extract cloudevents functions, privateize file-global identifiers

grant:
- Moves `getBinaryCloudEventContext` into `src/cloudevents.ts`
- Inlines global `rawBodySaver` method (not really needed outside `getServer`
- Cleans up a few comments


grant:
Last part of #134.

PR #139, opened 2020-05-14
Use CloudEvents v1.0 in CloudEventsContext and tests

hdp617:


test/invoker.ts
```diff
@@ -156,38 +156,51 @@ describe('CloudEvents request to event function', () => {
...
   }
+
+  const specversion = '1.0';
+  const type = 'com.google.cloud.storage';
+  const source =
+    'https://github.com/GoogleCloudPlatform/functions-framework-nodejs';
+  const subject = 'test-subject';
+  const id = 'test-1234-1234';
+  const time = '2020-05-13T01:23:45Z';
+  const dataschema =
```
grant, line 167: This isn't a true dataschema of the payload. We should just not include this CE attr.

hdp617: Removed!

PR #141, opened 2020-05-15
Fix PubSub Event Example

twistedpair:
Use the proper function name for the example.


twistedpair:
@googlebot I signed it!

grant:
Thanks!

Issue #22, opened 2019-04-21
process.on('exit', ... always errors


anishkny:
Doesn't the following line mean that the process will always end in an error?

https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/518f2d0b4f171273dc9e7e724f6a3ffdc90db8d9/src/invoker.ts#L531

What is the non-error-producing recommended way to stop the server? Say after testing in a CI environment?

grant:
That's a good question.
I don't think there is a great way to stop the server now without an error.

Do you have a recommendation for what could fix this? Simply removing the error?

grant:
I think we'd need to change the hander to return `0` and not call `logAndSendError`.

anishkny:
Hey sorry had forgotten about this. Something like this would be very useful:

```
npx @google-cloud/functions-framework --run-script "./test-my-api.sh"
```

This should spin up the API and then run `test-my-api.sh` (or whatever) and when the script is done, quit the API and marshal back the exit code for CI.

chrisLeeTW:
@grant I thinks this is very good and simple and elegant solutions for this problem.

grant:
This should be fixed now on `master` with https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/187.

Issue #143, opened 2020-05-22
Smooth Exit


grant:
The FF should smoothly exit, even with a process interrupt. This would create a better developer experience.

This is what a user sees when we CTRL-C:

```sh
Serving function...
Function: function
URL: http://localhost:8080/
^CReceived SIGINT
Error: Process exited with code 0
    at process.<anonymous> (<folder>/node_modules/@google-cloud/functions-framework/build/src/invoker.js:396:29)
    at process.emit (events.js:203:13)
    at process.EventEmitter.emit (domain.js:471:20)
    at process.exit (internal/process/per_thread.js:158:15)
    at Server.<anonymous> (<folder>/node_modules/@google-cloud/functions-framework/build/src/invoker.js:402:29)
    at Object.onceWrapper (events.js:291:20)
    at Server.emit (events.js:203:13)
    at Server.EventEmitter.emit (domain.js:471:20)
    at emitCloseNT (net.js:1572:8)
    at processTicksAndRejections (internal/process/task_queues.js:77:11)
```

Source:

```ts
process.on('exit', code => {
  logAndSendError(new Error(`Process exited with code ${code}`), latestRes);
});
```

We're not really doing any good by logging the stacktrace of the Node process, maybe a one-liner error would be better.

grant:
Confirmed, this is still an issue today.

`    "@google-cloud/functions-framework": "^1.6.0"`

grant:
This should be fixed now on `master` with https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/187.

@hdp617 maybe cut a minor version? We might want to cut a major version after the CE fix.

PR #144, opened 2020-05-22
fix: do not err on process exit

grant:
Fixes: #22 #143

This change the behavior the process exit event handler.
I don't know exactly why we have a `latestRes` modification here or if it's needed. I would assume a process exit does not need to send the latest HTTP response as an exit would not naturally happen.

## Before

```sh
Serving function...
Function: function
URL: http://localhost:8080/
^CReceived SIGINT
Error: Process exited with code 0
    at process.<anonymous> (<folder>/node_modules/@google-cloud/functions-framework/build/src/invoker.js:396:29)
    at process.emit (events.js:203:13)
    at process.EventEmitter.emit (domain.js:471:20)
    at process.exit (internal/process/per_thread.js:158:15)
    at Server.<anonymous> (<folder>/node_modules/@google-cloud/functions-framework/build/src/invoker.js:402:29)
    at Object.onceWrapper (events.js:291:20)
    at Server.emit (events.js:203:13)
    at Server.EventEmitter.emit (domain.js:471:20)
    at emitCloseNT (net.js:1572:8)
    at processTicksAndRejections (internal/process/task_queues.js:77:11)
```

## After

```sh
Serving function...
Function: function
URL: http://localhost:8080/
^CReceived SIGINT
Process exited with code 0
```

### DIFF
--- a/src/invoker.ts
+++ b/src/invoker.ts
@@ -259,7 +259,7 @@ export class ErrorHandler {
     });
 
     process.on('exit', code => {
-      logAndSendError(new Error(`Process exited with code ${code}`), latestRes);
+      console.log(`Process exited with code ${code}`);
     });
 
     ['SIGINT', 'SIGTERM'].forEach(signal => {
--- a/src/invoker.ts
+++ b/src/invoker.ts
@@ -258,7 +258,7 @@ export class ErrorHandler {
       logAndSendError(err, latestRes, killInstance);
     });
 
-    process.on('exit', code => {
+    process.on('SIGINT', code => {
       console.log(`Process exited with code ${code}`);
     });
 


src/invoker.ts
```diff
@@ -259,7 +259,7 @@ export class ErrorHandler {
     });
 
     process.on('exit', code => {
-      logAndSendError(new Error(`Process exited with code ${code}`), latestRes);
+      console.log(`Process exited with code ${code}`);
```
hdp617, line 262: The process could exit from the user's function. I think users would lose error logs in this case.

I think we should only only do this for `SIGINT`.

grant: Updated `exit` to `SIGINT`.

src/invoker.ts
```diff
@@ -258,8 +258,8 @@ export class ErrorHandler {
       logAndSendError(err, latestRes, killInstance);
     });
 
-    process.on('exit', code => {
-      logAndSendError(new Error(`Process exited with code ${code}`), latestRes);
+    process.on('SIGINT', code => {
+      console.log(`Process exited with code ${code}`);
     });
 
     ['SIGINT', 'SIGTERM'].forEach(signal => {
```
hdp617, line 265: `SIGINT` is also handled here. Could you please update this?

src/invoker.ts
```diff
@@ -258,8 +258,8 @@ export class ErrorHandler {
       logAndSendError(err, latestRes, killInstance);
     });
 
-    process.on('exit', code => {
```
hdp617, line 261: I think this would remove error logging when the process exits from user's function.

I do think we should not log and send error when process exits normally though. How about checking the status code and only `logAndSendError` for non-0 ones? Or handing `SIGINT` before `exit`?

hdp617:
@grant What's the status of this PR?

grant:
> @grant What's the status of this PR?

I don't know how to test this change, don't want to break GCF, and haven't continued with it.
Do you know the solution here?

grant:
This PR is outdated and fixed with https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/126.

PR #145, opened 2020-05-28
docs: remove incorrect pubsub docs

grant:
The current docs are incorrect. We don't have a replacement just yet.
See https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/96 https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/41


grant:
Pinged on Chat.

grant:
Yeah, we should add doc here when we get a correct guide.

PR #158, opened 2020-07-30
feat: add a script to run conformance tests

hdp617:


run-conformance-tests.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
grant, line 1: Can we put this file in `test` or somewhere besides top level?

PR #187, opened 2020-08-05
fix: do not send error when error code is fine

grant:
Doesn't log assuming an error on process.exit. Keeps same functionality.
Tested locally.

```
 npx @google-cloud/functions-framework --target=helloWorld

Serving function...
Function: helloWorld
URL: http://localhost:8080/
^CReceived SIGINT
```

Not sure about upstream impact.

### DIFF
--- a/src/invoker.ts
+++ b/src/invoker.ts
@@ -28,7 +28,7 @@ import * as http from 'http';
 import * as onFinished from 'on-finished';
 
 import {FUNCTION_STATUS_HEADER_FIELD} from './types';
-import {logAndSendError} from './logger';
+import {sendCrashResponse} from './logger';
 import {isBinaryCloudEvent, getBinaryCloudEventContext} from './cloudevents';
 import {
   HttpFunction,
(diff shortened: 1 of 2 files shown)


src/invoker.ts
```diff
@@ -305,16 +305,22 @@ export class ErrorHandler {
...
 
     process.on('unhandledRejection', err => {
       console.error('Unhandled rejection');
-      logAndSendError(err, latestRes, killInstance);
+      sendCrashResponse({err, res: latestRes, callback: killInstance});
     });
 
     process.on('exit', code => {
-      logAndSendError(new Error(`Process exited with code ${code}`), latestRes);
+      if (code) {
```
hdp617, line 317: With this check, it won't send an error response for code 0.

grant: That's the point. 0 represents a successful exit status, and thus should not produce an error.

What is expected?

hdp617: I _think_ it still needs to be reported to the supervisor.

grant: Oh, I see the error in my logic.
PTAL again.

Issue #13, opened 2019-04-10
Cloud Function in Typescript


paulwoelfelcpl:
The functions framework expect a index.js in the local directory. I'm writing my functions in Typescript and test it locally with nodemon, ts-node and a ts file, which starts the express server. 

Is there a better way than running `tsc` every time something changes and then start the functions-framework?

paulwoelfelcpl:
My solution right now is to use [tsc-watch](https://www.npmjs.com/package/tsc-watch) and the following command:

```bash
npx tsc-watch --onSuccess 'npx @google-cloud/functions-framework --function-target=main --port=8080 --function-signature-type=http'
```

In the local directory I created `index.js`, which loads the transpiled js:

```javascript
const cloudFunctions = require("./dist");

Object.keys(cloudFunctions).forEach(
  key => (exports[key] = cloudFunctions[key])
);

```

And that's the `tsconfig.json`

```json
{
  "compilerOptions": {
    "outDir": "./dist",
    "target": "es2018",
    "lib": ["dom", "es2018"],
    "module": "commonjs",
    "esModuleInterop": true,
    "pretty": true,
    "sourceMap": true
  },
  "include": ["./src/**/*"],
  "exclude": ["node_modules"]
}
```

stew-r:
Hey there -- I think your solution is a good one. The framework should work with standard JS tooling so I'm glad you were able to make this work.

QuantumInformation:
Why do the docs say to do `npm install @google-cloud/functions-framework`

but then tell you to use npx later?

stew-r:
They're two different tools. [See here](https://stackoverflow.com/questions/50605219/difference-between-npx-and-npm) for a more detailed explanation.

fernandofleury:
This is how I'm achieving it using `nodemon`:

```
"start": "functions-framework --target=event",
"build": "npm run build-ts",
"watch": "concurrently -k -p \"[{name}]\" -n \"TypeScript,Node\" -c \"yellow.bold,cyan.bold,green.bold\" \"npm run watch-ts\" \"npm run watch-node\"",
"watch-node": "nodemon --watch ./build/ --exec npm run start",
```

PR #192, opened 2020-08-09
docs: add ts example

naseemkullah:
Also lints README by indenting some codeblocks in enumerated lists.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -5,7 +5,7 @@ This directory contains advanced docs around the Functions Framework.
 - [Testing events and Pub/Sub](events.md)
 - [Debugging Functions](debugging.md)
 - [Running and Deploying Docker Containers](docker.md)
-- [Writing a function in Typescript](typescript.md)
+- [Writing a Function in Typescript](typescript.md)
 
 ## TODO Docs
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -5,6 +5,7 @@ This directory contains advanced docs around the Functions Framework.
 - [Testing events and Pub/Sub](events.md)
 - [Debugging Functions](debugging.md)
 - [Running and Deploying Docker Containers](docker.md)
+- [Writing a function in Typescript](typescript.md)
 
 ## TODO Docs
 
@@ -16,28 +17,28 @@ The Functions Framework works with standard tooling that you might use when writ
 
 1. Write an `index.js` file containing your Node.js function:
 
-```js
-exports.helloWorld = (req, res) => {
-  res.send('Hello, World');
-};
-```
+    ```js
+    exports.helloWorld = (req, res) => {
+      res.send('Hello, World');
+    };
+    ```
 
 2. Install the Functions Framework:
 
-```sh
-npm install @google-cloud/functions-framework
-```
+    ```sh
+    npm install @google-cloud/functions-framework
+    ```
 
 3. Run `node`, enable the inspector and run the Functions Framework:
 
-```sh
-node --inspect node_modules/@google-cloud/functions-framework --target=helloWorld
-...
-Debugger listening on ws://127.0.0.1:9229/5f57f5e9-ea4b-43ce-be1d-6e9b838ade4a
-For help see https://nodejs.org/en/docs/inspector
-Serving function...
-Function: helloWorld
-URL: http://localhost:8080/
-```
+    ```sh
+    node --inspect node_modules/@google-cloud/functions-framework --target=helloWorld
+    ...
+    Debugger listening on ws://127.0.0.1:9229/5f57f5e9-ea4b-43ce-be1d-6e9b838ade4a
+    For help see https://nodejs.org/en/docs/inspector
+    Serving function...
+    Function: helloWorld
+    URL: http://localhost:8080/
+    ```
 
 You can now use an IDE or other tooling to add breakpoints, step through your code and debug your function.
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -1,0 +1,53 @@
+# Typescript
+
+This guide shows you an example of how to develop a function in Typescript with
+the Functions Framework.
+
+1. Use [gts](https://github.com/google/gts) to configure Typescript.
+
+    ```sh
+    npx gts init
+    ```
+
+2. Install the required package(s):
+
+    ```sh
+    npm install @google-cloud/functions-framework
+    ```
+
+    In this example we are using an HTTP signature and require express
+    types:
+
+    ```sh
+    npm install @types/express -D
+    ```
+
+3. Add a `start` script to `package.json`, passing in the `--source` flag to
+   point to the compiled code directory:
+
+    ```js
+      "scripts": {
+        "start": "functions-framework --source=build/src/ --target=helloWorld",
+        ...
+      }
+    ```
+
+4. Create a `src/index.ts` file with the following contents:
+
+    ```ts
+    import type { HttpFunction } from '@google-cloud/functions-framework/build/src/functions';
+
+    export const helloWorld: HttpFunction = (req, res) => {
+      res.send('Hello, World');
+    };
+    ```
+
+5. Compile and start the built-in local development server:
+
+    ```sh
+    $ npm run compile && npm start
+    ...
+    Serving function...
+    Function: helloWorld
+    URL: http://localhost:8080/
+    ```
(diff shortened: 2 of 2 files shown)


docs/typescript.md
```diff
@@ -0,0 +1,53 @@
...
+    export const helloWorld: HttpFunction = (req, res) => {
+      res.send('Hello, World');
+    };
+    ```
+
+5. Compile and start the built-in local development server:
+
+    ```sh
+    $ npm run compile && npm start
+    ...
```
grant, line 49: - Where do we define the `compile` script?
- We probably want to separate the compiler and function starter. For example, we'd probably want to run `tsc --watch` or something similar in a `watch` script.
- We should not have the `$` symbol in the docs.

docs/typescript.md
```diff
@@ -0,0 +1,53 @@
...
+
+    In this example we are using an HTTP signature and require express
+    types:
+
+    ```sh
+    npm install @types/express -D
+    ```
+
+3. Add a `start` script to `package.json`, passing in the `--source` flag to
+   point to the compiled code directory:
```
grant, line 26: Where do we specify the compiled code directory? I think that step is missing from the instructions.

docs/typescript.md
```diff
@@ -0,0 +1,53 @@
...
+    ```sh
+    npm install @google-cloud/functions-framework
+    ```
+
+    In this example we are using an HTTP signature and require express
+    types:
+
+    ```sh
+    npm install @types/express -D
+    ```
```
grant, line 23: We could probably have simpler instructions:

```
npm i @google-cloud/functions-framework
npm i -D @types/express
```

docs/typescript.md
naseemkullah: re: The `compile` script, it is added by `gts`, should that be mentioned?

docs/typescript.md
naseemkullah: similarly, this is added by `gts` in `tsconfig.json`, so we just mention that?

docs/typescript.md
grant: oh, yeah maybe mention it to not confuse people. Something like:

> The `npm run compile` script is defined by `gts`.

docs/typescript.md
grant: No need to mention, thanks for telling me. Folks will figure out.

docs/typescript.md
naseemkullah: I'm still developing my first function with this framework and have not gotten around to using `tsc --watch` yet but will add it to the doc. Quick question though, is this used like `nodemon`, i.e. when developing only?

naseemkullah: Ok just read up and it is actually even defined as the `nodemon` for typescript, nice!

docs/typescript.md
```diff
@@ -0,0 +1,53 @@
...
+
+    export const helloWorld: HttpFunction = (req, res) => {
+      res.send('Hello, World');
+    };
+    ```
+
+5. Start the built-in local development server:
+
+    ```sh
+    $ npm run watch
```
naseemkullah, line 47: `$` here since we are displaying output below

naseemkullah:
> Thanks for the guide! Some requested changes:
> 
> * Can you brush up the instructions such that they can be copy-paste-able? I think there are some missing steps.
> * It might be worth adding a link to this comment: [#13 (comment)](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/13#issuecomment-481627025)
>   
>   * That extra feature of tsc-watch would be really useful.

Thanks for the feedback @grant, ready for another review at this point.

docs/README.md
```diff
@@ -5,6 +5,7 @@ This directory contains advanced docs around the Functions Framework.
 - [Testing events and Pub/Sub](events.md)
 - [Debugging Functions](debugging.md)
 - [Running and Deploying Docker Containers](docker.md)
+- [Writing a function in Typescript](typescript.md)
```
grant, line 8: nit: Uppercase **F**unction since this is a document title.

docs/typescript.md
grant: nit: Can we have the input and output separated so we can easily copy/paste? Something like:

---


  ```sh
  npm run watch
  ```

  This will continuously watch changes to your TypeScript project and recompile when changes are detected:

  ```
  [12:34:56 AM] Starting compilation in watch mode...
  ...
  Serving function...
  Function: helloWorld
  URL: http://localhost:8080/
  ```

---

You can tell, I really don't want the `$` symbol. And want to clearly separate stdin/stdout.

docs/typescript.md
```diff
@@ -0,0 +1,53 @@
...
+
+    ```js
+      "scripts": {
+        "start": "functions-framework --source=build/src/ --target=helloWorld",
+        "watch": "tsc-watch --onSuccess 'npm start'",
+        ...
+      }
+    ```
+
+4. Create a `src/index.ts` file with the following contents:
```
grant, line 34: Looks like `gts` already creates this file when I tried it.

Say "Replace" instead of "Create".

docs/typescript.md
```diff
@@ -0,0 +1,53 @@
...
+    ```
+
+3. Add a `start` script to `package.json`, passing in the `--source` flag to
+   point to the compiled code directory (configured by `gts` in this example).
+   Also add a `watch` script to use for development:
+
+    ```js
+      "scripts": {
+        "start": "functions-framework --source=build/src/ --target=helloWorld",
+        "watch": "tsc-watch --onSuccess 'npm start'",
```
grant, line 29: Unfortunately I wasn't able to get `tsc-watch` to properly refresh when changing code.

Luckily, I tried a different solution that works really well that I think we should document:

```
    "start": "functions-framework --source=build/src/ --target=helloWorld",
    "watch": "concurrently -k -p \"[{name}]\" -n \"TypeScript,Node\" -c \"yellow.bold,cyan.bold,green.bold\" \"npm run watch-ts\" \"npm run watch-node\"",
    "watch-node": "nodemon --watch ./build/ --exec npm run start",
    "watch-ts": "tsc -w"
```

This is the solution provided here: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/13#issuecomment-526628422

Here is a screenshot:

Screen Shot 2020-08-09 at 23 52 17

Was this watch working for you when changing code?

naseemkullah: `tsc-watch` does indeed work when code changes are made on my end that result in  successful compilation (`--onSuccess`)  🤔, do you mind trying. again?

```
    "watch": "tsc-watch --onSuccess 'npm start'",
```

docs/typescript.md
naseemkullah: done! I've also appended "in watch mode:" to the instruction on `5.` to disambiguate from a regular start, although the code blocked command already does.

docs/typescript.md
grant: I'm not seeing the Node process properly refresh. It sometimes does, but lags, or never refreshes.

Here's a GIF where you can see me refreshing and not seeing updates:
gif2

naseemkullah: Oh shucks ok, FWIW I've heard `ts-node` and `nodemon` can also achieve this as per https://stackoverflow.com/questions/37979489/how-to-watch-and-reload-ts-node-when-typescript-files-change

Any preference?

grant: When testing, I was able to get things working like above which uses `nodemon`: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/192#discussion_r467687210

I don't have a real preference so long as things work. The `nodemon` solution seems fine.

naseemkullah: ooc what does Google do internally for this today? Or are functions generally written in js?

grant: For serverless, we don't really have workflows where we need realtime updates to the FF. We don't dogfood Functions as much as I'd like. We're mostly using tsc --watch and running unit tests for the TS modules we have.

But this is important for external use of FFs and Cloud Functions / Cloud Run devX.

grant:
@naseemkullah Ping me when this is ready for review. I think the only thing is the `tsc-watch` part.

naseemkullah:
> @naseemkullah Ping me when this is ready for review. I think the only thing is the `tsc-watch` part.

Perfect, thanks @grant, should have it fixed up soon.

docs/typescript.md
naseemkullah: Alright I've settled on a nodemon+concurrently but with just 1 script instead of 3 as per https://fireship.io/snippets/hot-reload-node-typescript-server/

Issue #191, opened 2020-08-08
Requesting /favicon.ico URL invokes function, ignores its response, logs error to console


kkm000:
## Unexpected behavior

`invoker.ts` handles the URLs '/favicon.ico' and '/robots.txt' specially, by sending back a 404 response. I _assume_ that, by design, this was all what these two URLs were supposed to do, so that the behavior I am seeing is really a bug.

### Everything goes as expected...

1. Follow the [HelloWorld function tutorial](https://cloud.google.com/functions/docs/functions-framework). Confirm the response from `curl localhost:8080` at the last step.
2. Add a code line to the function to print a debug message to console as early as possible (the code should look as below), and run the server with `npm start` again.
```js
exports.helloWorld = (req, res) => {
  console.log('helloWorld is invoked');  /* Print a line every time the function is invoked */
  res.send('Hello, World');
};
```
3. Execute `curl -s http://localhost:8080`. Observe that, in addition to the reply being returned, a line `helloWorld is invoked` is printed to the server console.
4. Execute `curl -sv http://localhost:8080/favicon.ico`. Quite as expected, the command returns a `404 Not Found` response¹.
 \____________  
¹ And, somewhat less expected, courtesy of Express and on its conscience, also a text/plain body with the text "Not Found" which is not required, provides no extra information and thus unneeded, goes without the terminating `\r\n` as text/plain is supposed to be. But this is what Express does by default.

### ...until it does not

But much less than expected, the server console now shows:
```
helloWorld is invoked
Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at ServerResponse.setHeader (_http_outgoing.js:470:11)
    at ServerResponse.header (node_modules\express\lib\response.js:771:10)
    at ServerResponse.send (node_modules\express\lib\response.js:170:12)
    at ServerResponse.json (node_modules\express\lib\response.js:267:15)
    at ServerResponse.send (node_modules\express\lib\response.js:158:21)
    at helloWorld (index.js:3:9)
    at process.nextTick (node_modules\@google-cloud\functions-framework\build\src\invoker.js:100:17)
    at process._tickCallback (internal/process/next_tick.js:61:11)
```
This is rather bad. The function _is actually invoked_, with all its possible side effects. But whatever it returns is not seen by the caller: the framework has already sent a complete response to the HTTP trigger:
```
< HTTP/1.1 404 Not Found
< X-Powered-By: Express
[...]

PR #193, opened 2020-08-09
Don't call the function after a 404 on /{robots.txt,favicon.ico}

kkm000:
Other changes:
 * Send an empty response body for the same with the 404, since   these URLs are not intended for human consumption.
 * Add an invocation count assertion to the test to confirm that   these URLs do not in fact result in the function call.
 * Accept npm's lockfile update of this own package version.
 * Nit: wrap long comments at 80 characters.

Fixes: #191

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@google-cloud/functions-framework",
-  "version": "1.6.0",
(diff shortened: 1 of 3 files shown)


kkm000:
My pleasure, thanks! I need to learn the thing anyway. :)

Issue #206, opened 2020-09-01
Unable to remove response header


SDrinkwater:
There doesn't seem to be a way to remove a response header. I am trying to remove the X-Powered-By header. I am only able to set its value, through which I can effectively achieve the same result, but it would be nice to be able to actually remove the header.

Is there some way to achieve this?

Issue originally raised here: https://github.com/firebase/firebase-functions/issues/754

grant:
Hi @SDrinkwater,
Sorry, I don't understand your question. With `res.send(` in Express, you can send any response.

https://expressjs.com/en/api.html#res.send

The `X-Powered-By` is not coming from this library.

rhodgkins:
I'd also like to see this functionality added as we are also trying to remove the `x-powered-by` header.

@grant I think this header is coming from this library as it creates the express app:

https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/5d61008f7855cfc10877f6bf580eaef2487a4ca8/src/invoker.ts#L349

The default setting for `x-powered-by` is `true` which means the value of the header will be `express` ([see the docs](https://expressjs.com/en/api.html#app.settings.table)).

Then any sub-apps that are used for client function routes have their `x-powered-by` setting ignore.

You can try it with the following:

```js
const app = express()
app.get('/', (req, res) => {
   res.send('Will have x-powered-by header')
})    

const subapp = express()
subapp.disable('x-powered-by')
subapp.all('*', (req, res) => {
   res.send(`Will still have x-powered-by header despite being disabled: subapp.enabled('x-powered-by')=${subapp.enabled('x-powered-by')}`)
})
app.use('/subapp', subapp)

// Matches how functions are called
app.all('/calledsubapp', (req, res, next) => {
   subapp(req, res, next)
})
```

Querying the main app:
```
curl -i http://localhost:3000       
HTTP/1.1 200 OK
X-Powered-By: Express
Content-Type: text/html; charset=utf-8
Content-Length: 29
ETag: W/"1d-6q73dA/19m+t4qsFOySdg9LHLZI"
Date: Tue, 15 Sep 2020 08:09:26 GMT
Connection: keep-alive

Will have x-powered-by header     
```

Querying the sub app:
```
curl -i http://localhost:3000/subapp
HTTP/1.1 200 OK
X-Powered-By: Express
Content-Type: text/html; charset=utf-8
Content-Length: 96
[...]

PR #214, opened 2020-09-15
feat: add helmet express middleware

grant:
Fixes https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/206

- Adds https://www.npmjs.com/package/helmet middleware per Express' security best practices
  - https://expressjs.com/en/advanced/best-practice-security.html

CC: @rhodgkins @SDrinkwater

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1880,6 +1880,11 @@
       "integrity": "sha512-F/1DnUGPopORZi0ni+CvrCgHQ5FyEAHRLSApuYWMmrbSwoN2Mn/7k+Gl38gJnR7yyDZk6WLXwiGod1JOWNDKGw==",
       "dev": true
     },
+    "helmet": {
+      "version": "4.1.1",
+      "resolved": "https://registry.npmjs.org/helmet/-/helmet-4.1.1.tgz",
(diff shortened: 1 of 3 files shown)


hdp617:
`helmet` does more than removing the `X-Powered-By` header. It adds other headers like `X-Download-Options`, `X-Content-Type-Options`, etc. that are only applicable when running in a browser. Do we want to set them by default?

grant:
I don't think they're only applicable when invoking via a browser. For example Strict-Transport-Security enforcing https is useful for APIs (more relevant when the library is not hosted on GCF).

Another option would be to enable helmet features one by one.

If there's a concern about adding too many headers, we can go one by one and see if we want the feature.

hdp617:
Should this be up to users to use `helmet` in their functions?

For example, if I use `helmet` in my function and want to disable `frameguard`,
```javascript
const app = express();
app.use(
  helmet({
    frameguard: false,
  })
);
app.use(testFunction);

function testFunction(req, res) {
  res.send("Hello World");
}

exports.testFunction = app;
```
the framework doesn't preserve my settings and adds the `X-Frame-Options` header. This could be confusing?

PR #224, opened 2020-10-02
Use functions-framework-conformance action in conformance CI

hdp617:


.github/workflows/conformance.yml
```diff
@@ -9,21 +9,44 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x]
+        runtime: ['nodejs10', 'nodejs12']
```
grant, line 12: Why not 14?

.github/workflows/conformance.yml
```diff
@@ -9,21 +9,44 @@ jobs:
...
+    - name: Setup Go
+      uses: actions/setup-go@v2
       with:
-        go-version: '1.14'
-    - uses: actions/checkout@v2
-    - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v1
+        go-version: '1.13'
+    - name: Run HTTP conformance tests
+      uses: GoogleCloudPlatform/functions-framework-conformance/action@hdp617-patch-1
```
grant, line 21: Can we remove the patch?

grant (approved):

.github/workflows/conformance.yml
hdp617: Yes, I'll remove it after I make another conformance release.

.github/workflows/conformance.yml
hdp617: Done.

PR #237, opened 2020-11-20
feat: implement legacy to cloudevent conversion

hdp617:
This is the first half of the event conversions. This PR includes
* legacy to CE
* enable validate mapping in the conformance tests
* some refactoring for the CE to legacy conversions.

### DIFF
--- a/src/eventMapping.ts
+++ b/src/eventMapping.ts
@@ -112,14 +112,6 @@ export function getCloudEvent(req: express.Request): CloudEventsContext {
   const context = event.context;
   const data = event.data;
 
-  cloudevent = {
-    contenttype: CE_CONTENT_TYPE,
-    id: context.eventId,
-    specversion: CE_SPEC_VERSION,
-    time: context.timestamp,
-    data: data,
-  };
-
   // Determine CloudEvent type attribute.
   if (!context.eventType) {
     throw new Error('Unable to find background event type');
@@ -128,14 +120,16 @@ export function getCloudEvent(req: express.Request): CloudEventsContext {
 
   // Determine CloudEvent service attribute.
   if (!context.resource) {
-    throw new Error ('Unable to find background event resource');
+    throw new Error('Unable to find background event resource');
   }
   let ceSource: string;
   if (typeof context.resource === 'string') {
     // Resource is a raw path.
     // We need to determine the background event service from its type.
     let service = '';
-    for (const [bService, ceService] of Object.entries(BACKGROUND_TO_CLOUDEVENT_SERVICE_MAP)) {
+    for (const [bService, ceService] of Object.entries(
+      BACKGROUND_TO_CLOUDEVENT_SERVICE_MAP
+    )) {
       if (context.eventType.startsWith(bService)) {
         service = ceService;
         break;
@@ -150,6 +144,16 @@ export function getCloudEvent(req: express.Request): CloudEventsContext {
     const resource: CloudFunctionsResource = context.resource;
     ceSource = `//${resource.service}/${resource.name}`;
   }
+
+  cloudevent = {
+    contenttype: CE_CONTENT_TYPE,
+    id: context.eventId,
+    specversion: CE_SPEC_VERSION,
+    time: context.timestamp,
+    data: data,
+    source: ceSource,
+    type: ceType,
+  };
 
   return cloudevent;
 }
--- a/src/eventMapping.ts
+++ b/src/eventMapping.ts
@@ -28,7 +28,7 @@ import {
 /**
  * Mapping between background event types and CloudEvent types.
  */
-const typeBackgroundToCloudEvent: Record<string, string> = {
+const BACKGROUND_TO_CLOUDEVENT_EVENT_TYPE_MAP: Record<string, string> = {
   'google.pubsub.topic.publish':
     'google.cloud.pubsub.topic.v1.messagePublished',
   'providers/cloud.pubsub/eventTypes/topic.publish':
@@ -67,11 +67,11 @@ const typeBackgroundToCloudEvent: Record<string, string> = {
 /**
  * Mapping between background event services and CloudEvent services.
  */
-const serviceBackgroundToCloudEvent: Record<string, string> = {
+const BACKGROUND_TO_CLOUDEVENT_SERVICE_MAP: Record<string, string> = {
   'providers/cloud.firestore/': 'firestore.googleapis.com',
-  'providers/google.firebase.analytics/': 'firebase.googleapis.com',
-  'providers/firebase.auth/': 'firebase.googleapis.com',
-  'providers/google.firebase.database/': 'firebase.googleapis.com',
+  'providers/google.firebase.analytics/': 'firebaseauth.googleapis.com',
+  'providers/firebase.auth/': 'firebaseauth.googleapis.com',
+  'providers/google.firebase.database/': 'firebasedatabase.googleapis.com',
   'providers/cloud.pubsub/': 'pubsub.googleapis.com',
   'providers/cloud.storage/': 'storage.googleapis.com',
   'google.pubsub': 'pubsub.googleapis.com',
@@ -79,16 +79,16 @@ const serviceBackgroundToCloudEvent: Record<string, string> = {
 };
 
 // Default CloudEvent spec version.
-const ceSpecversion = '1.0';
+const CE_SPEC_VERSION = '1.0';
 // Default CloudEvent content type.
-const ceContentType = 'application/json';
+const CE_CONTENT_TYPE = 'application/json';
 
 /**
  * Get CloudEvent from the request object.
  * @param req Express request object.
  * @return CloudEvent object or null.
  */
-export function getCloudEvent(req: express.Request): CloudEventsContext | null {
+export function getCloudEvent(req: express.Request): CloudEventsContext {
   let cloudevent: CloudEventsContext;
 
   // Handle a CloudEvent in binary content mode.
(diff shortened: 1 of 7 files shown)


src/events.ts
```diff
@@ -0,0 +1,194 @@
+// Copyright 2020 Google LLC
```
grant, line 1: I think we want a more descriptive filename, like `eventMapping.ts`.

src/events.ts
```diff
@@ -0,0 +1,194 @@
...
+    return cloudevent;
+  }
+
+  // Handle a CloudEvent in structured content mode.
+  if (isStructuredCloudEvent(req)) {
+    cloudevent = req.body as CloudEventsContext;
+    return cloudevent;
+  }
+
+  console.log('Converting from background event to CloudEvent');
```
grant, line 107: No console logs in user code.

src/events.ts
```diff
@@ -0,0 +1,194 @@
...
+  const event = getBackgroundEvent(req);
+  if (event === null) {
```
grant, line 109: Use JS logic:

```js
if (!event) {
```

src/events.ts
```diff
@@ -0,0 +1,194 @@
...
+    console.error('Unable to extract background event');
```
grant, line 110: No console statements. Throw if needed.

src/events.ts
```diff
@@ -0,0 +1,194 @@
...
+    return null;
+  }
+
+  const context = event.context;
+  const data = event.data;
+
+  // Determine CloudEvent type attribute.
+  if (context.eventType === undefined) {
```
grant, line 118: `foo === undefined` should be `!foo`.

src/events.ts
```diff
@@ -0,0 +1,194 @@
...
+    console.error('Unable to find background event type');
+    return null;
```
grant, line 120: `return;` is `return null;`

src/events.ts
```diff
@@ -0,0 +1,194 @@
...
+import {
+  isBinaryCloudEvent,
+  isStructuredCloudEvent,
+  getBinaryCloudEventContext,
+} from './cloudevents';
+
+/**
+ * Mapping between background event types and CloudEvent types.
+ */
+const typeBackgroundToCloudEvent: Record<string, string> = {
```
grant, line 31: Use UPPER_CASE for constants.

`BACKGROUND_TO_CLOUDEVENT_EVENT_TYPE_MAP`

https://google.github.io/styleguide/jsguide.html#naming-constant-names

src/events.ts
```diff
@@ -0,0 +1,194 @@
...
+  'providers/google.firebase.database/eventTypes/ref.delete':
+    'google.firebase.database.document.v1.deleted',
+  'providers/cloud.storage/eventTypes/object.change':
+    'google.cloud.storage.object.v1.finalized',
+};
+
+/**
+ * Mapping between background event services and CloudEvent services.
+ */
+const serviceBackgroundToCloudEvent: Record<string, string> = {
```
grant, line 70: `BACKGROUND_TO_CLOUDEVENT_SERVICE`

src/events.ts
```diff
@@ -0,0 +1,194 @@
...
+    // Resource is a raw path.
+    // We need to determine the background event service from its type.
+    let service = '';
+    for (const bService in serviceBackgroundToCloudEvent) {
+      const ceService = serviceBackgroundToCloudEvent[bService];
+      if (context.eventType.startsWith(bService)) {
+        service = ceService;
+        break;
+      }
+    }
```
grant, line 140:
```ts
let service = '';
for (const [bService, ceService] of Object.entries(BACKGROUND_TO_CLOUDEVENT_EVENT_TYPE_MAP)) {
  if (context.eventType.startsWith(bService)) {
    service = ceService;
    break;
  }
}
```

src/functions.ts
```diff
@@ -62,7 +78,27 @@ export interface CloudFunctionsContext {
...
-  resource?: string;
+  resource?: string | CloudFunctionsResource;
+}
+
+/**
+ * The Cloud Functions resource object for the event.
+ *
+ * @link https://cloud.google.com/functions/docs/writing/background#function_parameters
+ */
+export interface CloudFunctionsResource {
```
grant, line 89: - I'm not seeing where this is defined.
- I think this only applies to background functions?

src/events.ts
```diff
@@ -0,0 +1,194 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as express from 'express';
+import {
+  BackgroundEvent,
+  CloudFunctionsContext,
+  CloudFunctionsResource,
+  CloudEventsContext,
```
grant, line 20: I know we did not author this in this PR, but I think we should change:

`CloudEventsContext`
to
`CloudEventContext`

as we're talking about a single CloudEvent.

src/invoker.ts
```diff
@@ -142,11 +143,14 @@ export function wrapCloudEventFunction(
         }
       }
     );
-    let cloudevent = req.body;
-    if (isBinaryCloudEvent(req)) {
-      cloudevent = getBinaryCloudEventContext(req);
-      cloudevent.data = req.body;
+
+    const event = getCloudEvent(req);
+    if (event === null) {
```
grant, line 148:
```ts
if (!event) {
```

src/invoker.ts
```diff
@@ -142,11 +143,14 @@ export function wrapCloudEventFunction(
...
+      sendResponse(undefined, new Error('Unable to get CloudEvent'), res);
```
grant, line 149: Say something more descriptive like:

```ts
new Error('Unable to parse CloudEvent from HTTP request')
```

grant (approved):

grant:
@hdp617 status?

src/functions.ts
hdp617: Yes, this only applies to background functions. It's defined in [here](https://cloud.google.com/functions/docs/writing/background#function_parameters) under `context.resource` as "The resource that emitted the event.".

src/events.ts
hdp617: Ack.

hdp617:
Grant, thanks for the review. I've addressed the comments. Since it's not a small change, would you mind taking another look before I merge? Thanks!

src/eventMapping.ts
```diff
@@ -149,16 +151,6 @@ export function getCloudEvent(req: express.Request): CloudEventsContext | null {
...
 
-  cloudevent = {
-    contenttype: ceContentType,
-    id: context.eventId,
-    specversion: ceSpecversion,
-    time: context.timestamp,
-    data: data,
-    source: ceSource,
-    type: ceType,
-  };
```
grant: I believe this introduces a bug.

- `cloudevent.source`
- `cloudevent.type`

...are never set now.

If this bug is not caught by the test, we should add tests to ensure this is caught..

src/eventMapping.ts
```diff
@@ -104,43 +104,45 @@ export function getCloudEvent(req: express.Request): CloudEventsContext | null {
...
   const context = event.context;
   const data = event.data;
 
+  cloudevent = {
+    contenttype: CE_CONTENT_TYPE,
+    id: context.eventId,
+    specversion: CE_SPEC_VERSION,
+    time: context.timestamp,
+    data: data,
+  };
```
grant: The logic of `cloudevent` in this function, `getCloudEvent` is confusing because we have `cloudevent` as a let and we have multiple returns..

I think we should try to remove `cloudevent`. Keeping multiple returns is fine, but could be shorter. Example:

```ts
export function getCloudEvent(req: express.Request): CloudEventsContext {
  // Handle a CloudEvent in binary content mode.
  if (isBinaryCloudEvent(req)) {
    return {
      ...getBinaryCloudEventContext(req),
      data: req.body
    };
  }

  // Handle a CloudEvent in structured content mode.
  if (isStructuredCloudEvent(req)) {
    return req.body as CloudEventsContext;
  }

  const event = getBackgroundEvent(req);
  if (!event) {
    throw new Error('Unable to extract background event');
  }

```

src/eventMapping.ts
```diff
@@ -104,43 +104,45 @@ export function getCloudEvent(req: express.Request): CloudEventsContext | null {
...
+    throw new Error('Unable to find background event type');
   }
-  const ceType = typeBackgroundToCloudEvent[context.eventType];
+  const ceType = BACKGROUND_TO_CLOUDEVENT_EVENT_TYPE_MAP[context.eventType];
 
   // Determine CloudEvent service attribute.
-  if (context.resource === undefined) {
-    console.error('Unable to find background event resource');
-    return null;
+  if (!context.resource) {
```
grant, line 122: All of the property validation checks could be next to each other where possible. Comments shouldn't be needed given we have an error message.

```ts
if (!context.resource) throw new Error('Unable to find background event resource');
if (!context.eventType) throw new Error('Unable to find background event type');
...
```

src/eventMapping.ts
```diff
@@ -104,43 +104,45 @@ export function getCloudEvent(req: express.Request): CloudEventsContext | null {
...
+
   // Determine CloudEvent type attribute.
-  if (context.eventType === undefined) {
-    console.error('Unable to find background event type');
-    return null;
+  if (!context.eventType) {
+    throw new Error('Unable to find background event type');
   }
-  const ceType = typeBackgroundToCloudEvent[context.eventType];
+  const ceType = BACKGROUND_TO_CLOUDEVENT_EVENT_TYPE_MAP[context.eventType];
```
grant, line 119: Constant not used.

src/eventMapping.ts
```diff
@@ -104,43 +104,45 @@ export function getCloudEvent(req: express.Request): CloudEventsContext | null {
     return cloudevent;
   }
 
-  console.log('Converting from background event to CloudEvent');
   const event = getBackgroundEvent(req);
-  if (event === null) {
-    console.error('Unable to extract background event');
-    return null;
+  if (!event) {
+    throw new Error('Unable to extract background event');
```
grant, line 109: More specific `Error` classes would be nice. I think `TypeError` makes sense here.

grant (changes requested):
Thanks.

I believe there is a bug that we need to address (and test) and CI errors that need fixing before merging.

grant:
We'll use this as a good start in a different PR. We can't use this PR as the author isn't working on this product currently.

PR #242, opened 2020-12-24
fix: Do not pass numeric arguments to res.send

### LINKED ISSUE #239: express deprecated res.send(status): Use res.sendStatus(status) instead
We are getting an Unhandled rejection, whose error seem to be the integer `2`, which furthers triggers another error as shown below.

The issue is because in the logger https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/240defc0a6c7076fcffef89bdf6b95a0a1f95f48/src/logger.ts#L47 you are sending the data as is. However, for safety it would be better to first cast that to a string, in order to not trigger the logic https://github.com/expressjs/express/blob/508936853a6e311099c9985d4c11a4b1b8f6af07/lib/response.js#L130-L139

Or at least check if the data is a number and if so, convert it to string.

```
Unhandled rejection
2
Thu, 10 Dec 2020 15:23:12 GMT express deprecated res.send(status): Use res.sendStatus(status) instead at ../layers/google.nodejs.functions-framework/functions-framework/node_modules/@google-cloud/functions-framework/build/src/logger.js:34:13
Uncaught exception
RangeError [ERR_HTTP_INVALID_STATUS_CODE]: Invalid status code: 2 at ServerResponse.writeHead (_http_server.js:248:11) at ServerResponse._implicitHeader (_http_server.js:239:8) at ServerResponse.end (_http_outgoing.js:756:10) at /workspace/node_modules/@opentelemetry/plugin-http/build/src/http.js:212:87 at HttpPlugin._safeExecute (/workspace/node_modules/@opentelemetry/plugin-http/build/src/http.js:296:20) at ServerResponse.response.end (/workspace/node_modules/@opentelemetry/plugin-http/build/src/http.js:212:49) at ServerResponse.send (/layers/google.nodejs.functions-framework/functions-framework/node_modules/express/lib/response.js:221:10) at Object.sendCrashResponse (/layers/google.nodejs.functions-framework/functions-framework/node_modules/@google-cloud/functions-framework/build/src/logger.js:34:13) at process.<anonymous> (/layers/google.nodejs.functions-framework/functions-framework/node_modules/@google-cloud/functions-framework/build/src/invoker.js:271:22) at process.emit (events.js:315:20) at process.EventEmitter.emit (domain.js:483:12) at processPromiseRejections (internal/process/promises.js:209:33) at processTicksAndRejections (internal/process/task_queues.js:98:32)
_http_server.js:248
 throw new ERR_HTTP_INVALID_STATUS_CODE(originalStatusCode);
 ^
RangeError [ERR_HTTP_INVALID_STATUS_CODE]: Invalid status code: 2
 at ServerResponse.writeHead (_http_server.js:248:11)
 at ServerResponse._implicitHeader (_http_server.js:239:8)
 at write_ (_http_outgoing.js:642:9)
 at ServerResponse.end (_http_outgoing.js:753:5)
 at ServerResponse.send (/layers/google.nodejs.functions-framework/functions-framework/node_modules/express/lib/response.js:221:10)
 at Object.sendCrashResponse (/layers/google.nodejs.functions-framework/functions-framework/node_modules/@google-cloud/functions-framework/build/src/logger.js:34:13)
 at process.<anonymous> (/layers/google.nodejs.functions-framework/functions-framework/node_modules/@google-cloud/functions-framework/build/src/invoker.js:267:22)
 at process.emit (events.js:327:22)
 at process.EventEmitter.emit (domain.js:483:12)
 at process._fatalException (internal/process/execution.js:165:25) {
 code: 'ERR_HTTP_INVALID_STATUS_CODE'
```

sk-:
Fixes #239.

res.send when receiving a sole numeric argument will set thee statuc code to that number.

This breaks HTTP communication as some numbers are not valid HTTP response codes.

For example, in the original issue the code is being set to 2. Which probably comes from the UNKNOWN error response of an RPC call.

This makes that the function is reported as having a `connection error`.

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -44,7 +44,10 @@ export function sendCrashResponse({
   // execution sends the response between the check and 'send' call below.
   if (res && !res.headersSent) {
     res.set(FUNCTION_STATUS_HEADER_FIELD, 'crash');
-    res.send(err.message || err);
+    const message = err.message || err;
+    const sanitizedMessage =
+      typeof message === 'number' ? String(message) : message;
+    res.send(sanitizedMessage);
   }
   if (callback) {
     callback();
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -44,10 +44,7 @@ export function sendCrashResponse({
   // execution sends the response between the check and 'send' call below.
   if (res && !res.headersSent) {
     res.set(FUNCTION_STATUS_HEADER_FIELD, 'crash');
-    const message = err.message || err;
-    const sanitizedMessage =
-      typeof message === 'number' ? String(message) : message;
-    res.send(sanitizedMessage);
+    res.send((err.message || err) + '');
   }
   if (callback) {
     callback();


src/logger.ts
```diff
@@ -44,7 +44,10 @@ export function sendCrashResponse({
   // execution sends the response between the check and 'send' call below.
   if (res && !res.headersSent) {
     res.set(FUNCTION_STATUS_HEADER_FIELD, 'crash');
-    res.send(err.message || err);
+    const message = err.message || err;
+    const sanitizedMessage =
+      typeof message === 'number' ? String(message) : message;
+    res.send(sanitizedMessage);
```
grant, line 50:
```ts
// Send string response
res.send((err.message || err) +'');
```

grant (changes requested):
Sorry for the delay in reviewing. To fix the error, I suggest the logic be a bit simpler – we convert the crash response to always be a string.

src/logger.ts
sk-: Done.

grant (approved):

grant:
Will be published in next release of the FF. If there's urgency to publish this, ping here.

Jkudjo:
found this useful in my application. Gracias

Issue #198, opened 2020-08-14
Implicit Request interface extension causes error property rawBody does not exists on type


dobromyslov:
Cloud functions implicitly extend express Request interface with rawBody (see https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/src/invoker.ts#L44):

```
// We optionally annotate the express Request with a rawBody field.
// Express leaves the Express namespace open to allow merging of new fields.
declare global {
  // eslint-disable-next-line @typescript-eslint/no-namespace
  namespace Express {
    export interface Request {
      rawBody?: Buffer;
    }
  }
}
```

I get error ` Property 'rawBody' does not exist on type 'Request<ParamsDictionary>'` when implement handler which uses Request.rawBody attribute like below:

```
import {Response, Request} from 'express';

export class HandlerClass{
  public async http(request: Request, response: Response): Promise<void> {
    const body = request.rawBody;
    ...
  }
}
```

As a workaround I have to create and use my own type instead of native Express Request:
```
import * as express from 'express';

export interface Request extends express.Request {
  rawBody: Buffer;
}
```

If you have a look at the `firebase-functions` repository then you will see they use explicit extension of Request interface and then use it in code (https://github.com/firebase/firebase-functions/blob/master/src/providers/https.ts#L33)

I think `functions-framework-nodejs` should change implicit interface extension to explicit and provide stand-alone Request with rawBody attribute to avoid this workaround mess in depended applications code.

grant:
Hey @dobromyslov, yeah that does look like an error with the TS interfaces.

I believe this workaround was just to fix the TS compiler, I don't think we really expected anybody to use our `Request` interface.

Want to send a PR, just changing the interfaces?

dobromyslov:
Yes, I will prepare PR and send it in a couple of days.

I implemented custom handling of rawBody in TypeScript and found this issue.

PR #263, opened 2021-03-31
fix: remove custom rawBody interface from request invoker

grant:
Removes the added `rawBody` field from requests.

This is used in the body-parser [`verify`](https://www.npmjs.com/package/body-parser#verify) callback hook (it doesn't actually verify the payload).

I'm not sure exactly the history around this field. The field, `req.rawBody` is there when testing in GCF, and it stores the JSON HTTP POST body in a byte array.

Example request:

```
 curl -XPOST https://us-central1-....cloudfunctions.net/function-26 -H "content-type: application/json" -d '{ "name": "fo" }'
```

Example response (that you need to parse):

```
JSON.parse("{\"type\":\"Buffer\",\"data\":[123,32,34,110,97,109,101,34,58,32,34,102,111,34,32,125]}")
```

Fixes https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/198

This would be a breaking change.

### DIFF
--- a/src/invoker.ts
+++ b/src/invoker.ts
@@ -35,17 +35,6 @@ import {
   CloudEventFunctionWithCallback,
 } from './functions';
 
-// We optionally annotate the express Request with a rawBody field.
-// Express leaves the Express namespace open to allow merging of new fields.
-declare global {
-  // eslint-disable-next-line @typescript-eslint/no-namespace
-  namespace Express {
-    export interface Request {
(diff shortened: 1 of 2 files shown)


grant:
I think we actually just need to document the reason for this a bit more and perhaps update the interface.

The reason for this is described here:

https://flaviocopes.com/express-get-raw-body/

PR #272, opened 2021-04-07
Fix local development with Pub/Sub emulator

### LINKED ISSUE #41: Cloud events incorrectly unmarshalled: first argument is undefined, 2nd argument differs from production

### LINKED ISSUE #96: Event parsing doesn't match Cloud Functions in background mode

matthewrobertson:
Currently, deploying a function with `--trigger-topic` creates a Pub/Sub subscription that delivers events to an internal endpoint that marshals request payloads before forwarding them to a user's function. Unfortunately, this breaks local development with the Pub/Sub emulator as the Functions Frameworks are dependent on this request marshalling logic. 

In order to better support local development and testing with the Pub/Sub emulator, this commit introduces the event marshalling logic to the Functions Framework via express middleware that is enable when the `PUBSUB_EMULATOR_HOST` is set.

fixes #41, fixes #96

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -1,0 +1,131 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+import {Request, Response, NextFunction} from 'express';
+
+const PUBSUB_EVENT_TYPE = 'google.pubsub.topic.publish';
+const PUBSUB_MESSAGE_TYPE =
+  'type.googleapis.com/google.pubsub.v1.PubsubMessage';
+const PUBSUB_SERVICE = 'pubsub.googleapis.com';
+
+interface RawPubSubBody {
+  subscription: string;
+  message: {
+    data: string;
+    messageId: string;
+    attributes: {[key: string]: string};
+  };
+}
+
+interface PubSubEventBody {
+  context: {
+    eventId: string;
+    timestamp: string;
+    eventType: typeof PUBSUB_EVENT_TYPE;
+    resource: {
+      service: typeof PUBSUB_SERVICE;
+      type: typeof PUBSUB_MESSAGE_TYPE;
+      name: string | null;
+    };
+  };
+  data: {
+    '@type': typeof PUBSUB_MESSAGE_TYPE;
+    data: string;
+    attributes: {[key: string]: string};
+  };
+}
+
+interface RawPubSubRequest extends Request {
+  body: RawPubSubBody;
+}
+
+interface MarshalledPubSubRequest extends Request {
+  body: PubSubEventBody;
+}
+
+/**
+ * Type predicate that checks if a given Request is a RawPubSubRequest
+ * @param request a Request object to typecheck
+ * @returns true if this Request is a RawPubSubRequest
+ */
+const isRawPubSubRequest = (request: Request): request is RawPubSubRequest => {
+  return !!(
+    request.body &&
+    request.body.subscription &&
+    request.body.message &&
+    request.body.message.data &&
+    request.body.message.messageId
+  );
+};
+
+/**
+ * Extract the Pub/Sub topic name from the HTTP request path.
+ * @param path the URL path of the http request
+ * @returns the Pub/Sub topic name if the path matches the expected format,
+ * null otherwise
+ */
+const extractTopic = (path: string): string | null => {
+  const parsedTopic = path.match(/projects\/[^/?]+\/topics\/[^/?]+/);
+  if (parsedTopic) {
+    return parsedTopic[0];
+  }
+  console.warn('Failed to extract the topic name from the URL path.');
+  console.warn(
+    "Configure your subscription's push endpoint to use the following path: ",
+    'projects/PROJECT_NAME/topics/TOPIC_NAME'
+  );
+  return null;
+};
+
+/**
+ * Marshal the body of an HTTP request from a Pub/Sub subscription
+ * @param req an unmarshalled http request from a Pub/Sub push subscription
+ * @returns the marshalled request body expected by wrapEventFunction
+ */
+const marshalPubSubRequestBody = (req: RawPubSubRequest): PubSubEventBody => ({
+  context: {
+    eventId: req.body.message.messageId,
+    timestamp: new Date().toISOString(),
+    eventType: PUBSUB_EVENT_TYPE,
+    resource: {
+      service: PUBSUB_SERVICE,
+      type: PUBSUB_MESSAGE_TYPE,
+      name: extractTopic(req.path),
+    },
+  },
+  data: {
+    '@type': PUBSUB_MESSAGE_TYPE,
+    data: req.body.message.data,
+    attributes: req.body.message.attributes,
+  },
+});
+
+/**
+ * Express middleware used to marshal the HTTP request body received directly from a
+ * Pub/Sub subscription into the format that is expected downstream by wrapEventFunction
+ * @param req express request object
+ * @param res express response object
+ * @param next function used to pass control to the next middle middleware function in the stack
+ */
+export const legacyPubSubEventMiddleware = (
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  if (isRawPubSubRequest(req)) {
+    ((req as unknown) as MarshalledPubSubRequest).body = marshalPubSubRequestBody(
+      req
+    );
+  }
+  next();
+};
--- a/src/pubsub_middleware.ts
+++ b/src/pubsub_middleware.ts
@@ -18,39 +18,50 @@ const PUBSUB_MESSAGE_TYPE =
   'type.googleapis.com/google.pubsub.v1.PubsubMessage';
 const PUBSUB_SERVICE = 'pubsub.googleapis.com';
 
+/**
+ * The request body of an HTTP request received directly from a Pub/Sub subscription.
+ *
+ * @link https://cloud.google.com/pubsub/docs/push?hl=en#receiving_messages
+ */
 interface RawPubSubBody {
+  /**
+   * The name of the subscription for which this request was made. Format is:
+   * projects/{project}/subscriptions/{sub}.
+   */
   subscription: string;
+  /**
+   * A message that is published by publishers and consumed by subscribers. The message must
+   * contain either a non-empty data field or at least one attribute.
+   *
+   * @link https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage
+   */
   message: {
+    /**
+     * Attributes for this message. If this field is empty, the message must contain non-empty
+     * data.
+     */
+    attributes?: {[key: string]: string};
+    /**
+     * The message data field. If this field is empty, the message must contain at least one
+     * attribute.
+     */
     data: string;
+    /**
+     * ID of this message, assigned by the server when the message is published. Guaranteed to be
+     * unique within the topic.
+     */
     messageId: string;
-    attributes: {[key: string]: string};
+    /**
+     * If non-empty, identifies related messages for which publish order should be respected. This
+     * field is not set by the Pub/Sub emulator.
+     */
+    orderingKey?: string;
+    /**
+     * The time at which the message was published, formatted as timestamp in RFC3339 UTC "Zulu"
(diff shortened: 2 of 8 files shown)


test/middleware.ts
```diff
@@ -0,0 +1,74 @@
...
+        },
+        context: {
+            eventId: '1', 
+            eventType: 'google.pubsub.topic.publish', 
+            resource: {
+               name: 'projects/FOO/topics/BAR_TOPIC', 
+               service: 'pubsub.googleapis.com', 
+               type: 'type.googleapis.com/google.pubsub.v1.PubsubMessage'
+            },
+            timestamp: new Date().toISOString(),
```
matthewrobertson, line 61: This will cause flaky tests. We need to add a dependency to freeze the clock in unit tests.

src/server.ts
```diff
@@ -97,6 +98,13 @@ export function getServer(
   // http://expressjs.com/en/advanced/best-practice-security.html#at-a-minimum-disable-x-powered-by-header
   app.disable('x-powered-by');
 
+  // If a Pub/Sub subscription is configured to invoke a user's function directly, the request body
+  // needs to be marshalled into the structure that wrapEventFunction expects. This unblocks local
+  // development with the Pub/Sub emulator
+  if (functionSignatureType === SignatureType.EVENT) {
```
grant, line 104: Don't we need to add `PUBSUB_EMULATOR_HOST` check somewhere?

src/middleware.ts
```diff
@@ -0,0 +1,131 @@
...
+interface RawPubSubBody {
+  subscription: string;
+  message: {
+    data: string;
+    messageId: string;
+    attributes: {[key: string]: string};
+  };
+}
+
+interface PubSubEventBody {
```
grant, line 30: - Can we put interfaces in `src/functions.ts` with our other interfaces?
- Can we partially re-use / extend an existing interface?

src/middleware.ts
```diff
@@ -0,0 +1,131 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+import {Request, Response, NextFunction} from 'express';
+
+const PUBSUB_EVENT_TYPE = 'google.pubsub.topic.publish';
+const PUBSUB_MESSAGE_TYPE =
+  'type.googleapis.com/google.pubsub.v1.PubsubMessage';
+const PUBSUB_SERVICE = 'pubsub.googleapis.com';
+
+interface RawPubSubBody {
```
grant, line 21: nit: Can you add a comment for every interface?

src/middleware.ts
```diff
@@ -0,0 +1,131 @@
+// Copyright 2019 Google LLC
```
grant, line 1: nit: Can we be a bit more specific with this file? I'd recommend `middlewarePubSub.ts`, `middlewareHelper.ts`, or something similar. We define middleware elsewhere in the repo and may have unrelated to Pub/Sub middleware in the future.

src/middleware.ts
```diff
@@ -0,0 +1,131 @@
...
+  );
+};
+
+/**
+ * Extract the Pub/Sub topic name from the HTTP request path.
+ * @param path the URL path of the http request
+ * @returns the Pub/Sub topic name if the path matches the expected format,
+ * null otherwise
+ */
+const extractTopic = (path: string): string | null => {
```
grant, line 77: nit: rename to `extractPubSubTopic`

src/middleware.ts
```diff
@@ -0,0 +1,131 @@
...
+interface MarshalledPubSubRequest extends Request {
+  body: PubSubEventBody;
+}
+
+/**
+ * Type predicate that checks if a given Request is a RawPubSubRequest
+ * @param request a Request object to typecheck
+ * @returns true if this Request is a RawPubSubRequest
+ */
+const isRawPubSubRequest = (request: Request): request is RawPubSubRequest => {
```
grant, line 61: I'd imagine the return type to be `boolean`.

src/middleware.ts
```diff
@@ -0,0 +1,131 @@
...
+  };
+}
+
+interface RawPubSubRequest extends Request {
+  body: RawPubSubBody;
+}
+
+interface MarshalledPubSubRequest extends Request {
+  body: PubSubEventBody;
+}
```
grant, line 54: Can we comment as to what these interfaces do / mean?

src/middleware.ts
```diff
@@ -0,0 +1,131 @@
...
+};
+
+/**
+ * Marshal the body of an HTTP request from a Pub/Sub subscription
+ * @param req an unmarshalled http request from a Pub/Sub push subscription
+ * @returns the marshalled request body expected by wrapEventFunction
+ */
+const marshalPubSubRequestBody = (req: RawPubSubRequest): PubSubEventBody => ({
+  context: {
+    eventId: req.body.message.messageId,
```
grant, line 97: Do we want to a more null safe access?

`req?.body?.message?.messageId`

We might get a bad body.

src/middleware.ts
```diff
@@ -0,0 +1,131 @@
...
+ * @param res express response object
+ * @param next function used to pass control to the next middle middleware function in the stack
+ */
+export const legacyPubSubEventMiddleware = (
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  if (isRawPubSubRequest(req)) {
+    ((req as unknown) as MarshalledPubSubRequest).body = marshalPubSubRequestBody(
```
grant, line 126: Can we add a comment to explain the type casting?

grant (changes requested):
Thanks!

Do you mind adding a little more detailed description in the PR for how you tested this locally?

For example, I don't see the use of `PUBSUB_EMULATOR_HOST` in the PR here. Some light instructions for how a user can test this manually would help. Even better would be to update [the docs we have on this content](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/docs/events.md) with the new setup.

src/middleware.ts
matthewrobertson: The Typechecker is protecting against this. The type predicate `isRawPubSubRequest` validates that all these fields are not undefined and that information is carried through the type system.

src/middleware.ts
matthewrobertson: Maybe this isn't the most clear way to write this. I will update it.

src/server.ts
matthewrobertson: You are right, my original idea was that this would check `process.env['PUBSUB_EMULATOR_HOST'] != null` here, but I wanted to put this up as an idea after thinking about @di's comments offline. Here we just treat this as any event conversion, where we detect the payload structure does not match the requested signature type and marshal it. This removes friction from the dev workflow: it just works out of the box without additional setup. It also improves portability of event functions (I can point my Pub/Sub push subscription at a function running in on any server with this change). The cost is some risk that we marshal an even that we aren't supposed to, but I think the benefits outweigh the risk as the current behaviour is clearly broken.

What do you think?

src/middleware.ts
matthewrobertson: I can update this. My thinking was eventually when we have more middleware, we would create `middleware/index.ts` and then imports for this module would stay the same...

src/middleware.ts
matthewrobertson: One bit I am confused about, this interface is a private implementation detail of the logic in this module. My understanding is the stuff in `functions.ts` is that it is public for clients of the framework.

src/pubsub_middleware.ts
```diff
@@ -0,0 +1,144 @@
...
+  return null;
+};
+
+/**
+ * Marshal the body of an HTTP request from a Pub/Sub subscription
+ * @param body an unmarshalled http request body from a Pub/Sub push subscription
+ * @param path the HTTP request path
+ * @returns the marshalled request body expected by wrapEventFunction
+ */
+const marshalPubSubRequestBody = (body: RawPubSubBody, path: string) => ({
```
grant, line 109: Do you mind annotating the return type for this function?

```ts
(body: RawPubSubBody, path: string): ??? => ({
```

test/pubsub_middleware.ts
```diff
@@ -0,0 +1,102 @@
...
+  message: {
+    data: 'eyJmb28iOiJiYXIifQ==',
+    messageId: '1',
+    attributes: {
+      test: '123'
+    }
+  }
+};
+
+const marshalledPubSubBody = (topic: string | null) => ({
```
grant, line 31: nit: Would be nice to have return type.

src/server.ts
di: (This makes sense to me!)

grant: Hmm.

Isn't the supervisor modifying GCF payloads such that GCF Pub/Sub prod works fine but FF Pub/Sub local doesn't? As long as we're sure that this version of the FF will work with prod, I suppose we're fine.

Perhaps `isRawPubSubRequestBody` evals to false in GCF.

src/middleware.ts
grant: For interfaces that developers will use to re-create HTTP POST request event payloads, they should be exported in the `functions.ts`. Otherwise, if private – no user will ever import it – then okay to keep private.

Public interfaces are kept in that file and exposed as stated here: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/package.json#L10

src/server.ts
matthewrobertson: > Perhaps `isRawPubSubRequestBody` evals to false in GCF.

Yes exactly, if the supervisor has already marshalled the payload, the request body will be different, and this middleware will be a no-op.

src/middleware.ts
matthewrobertson: OK I think it makes sense to keep this one private then.

PR #274, opened 2021-04-21
chore: add useful npm scripts

grant:
- Adds some useful scripts to this package: testing an individual test, tsc watching

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,6 @@
   },
   "scripts": {
     "test": "mocha build/test",
-    "test-individual": "mocha build/test -g 'loading function'",
     "check": "gts check",
     "clean": "gts clean",
     "compile": "tsc -p .",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -16,10 +16,12 @@
   },
   "scripts": {
     "test": "mocha build/test",
+    "test-individual": "mocha build/test -g 'loading function'",
```
matthewrobertson, line 19: Did you mean to hardcode the `'loading function'` here?

matthewrobertson: FYI this is pretty easy to do today with `npm run test -- -g 'my test'` so I am not sure it is worth adding a script for this

matthewrobertson (approved):
See comment about `test-individual`, otherwise LGTM

package.json
grant: Sounds good. I wanted to document how to do this somewhere.

I removed the script but added a note on how to do this in the contributing section.

PR #276, opened 2021-04-21
feat: updates event and cloudevent interfaces

grant:
- Updates the function interfaces for the Node FF
  - Adds a `LegacyEvent` interface
  - Updates `CloudEvent` interface to CE 1.0 (must have been <1.0 with `schemaurl`)

### DIFF
--- a/src/functions.ts
+++ b/src/functions.ts
@@ -41,6 +41,19 @@ export type HandlerFunction =
   | CloudEventFunctionWithCallback;
 
 /**
+ * A legacy event.
+ */
+export interface LegacyEvent {
+  data: object;
+  context: CloudFunctionsContext;
+}
+
+interface Data {
+  data: object;
+}
+export type LegacyCloudFunctionsContext = CloudFunctionsContext | Data;
+
+/**
  * The Cloud Functions context object for the event.
  *
  * @link https://cloud.google.com/functions/docs/writing/background#function_parameters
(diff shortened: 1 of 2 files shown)


src/functions.ts
```diff
@@ -62,7 +75,7 @@ export interface CloudFunctionsContext {
   /**
    * The resource that emitted the event.
    */
-  resource?: string;
+  resource?: string | object;
```
matthewrobertson, line 78: 👍 I noticed this was an issue in my last PR but was reluctant to touch it...

src/functions.ts
```diff
@@ -40,6 +40,19 @@ export type HandlerFunction =
...
+ */
+export interface LegacyEvent {
+  data: object;
+  context: CloudFunctionsContext;
+}
+
+interface Data {
+  data: object;
+}
+export type LegacyCloudFunctionsContext = CloudFunctionsContext | Data;
```
matthewrobertson, line 54: Is the `|` here because there are scenarios where the context argument to a user's function a `Data`?

matthewrobertson (approved):
LGTM

src/functions.ts
grant: 👌🏻

src/functions.ts
grant: Correct, iirc there were certain cases where the context argument was a data object.

PR #315, opened 2021-08-12
Functions ESM Sample

grant:
## Functions ESM Sample

Adds a sample showing how to use ES modules within Google Cloud Functions for the Node.js 12+ runtime.

Per internal discussion, keeping this sample in this repo rather than https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/2292.


matthewrobertson (approved):
LGTM

Issue #216, opened 2020-09-20
CloudFunctionsContext no longer exported?


techieshark:
I had working code as below on functions-framework@1.4.0:

```
import { CloudFunctionsContext } from '@google-cloud/functions-framework';
import { Message } from '@google-cloud/pubsub';

/**
 * A cloud function triggered by a PubSub message.
 * @see https://cloud.google.com/functions/docs/writing/background#function_parameters
 */
exports.pubsub = async function pubsub(
  pubsubMessage: Message,
  context: CloudFunctionsContext
) {
...
```

after updating to latest (1.7.1), I got TS error on first line:
```
Module '"../node_modules/@google-cloud/functions-framework/build/src/invoker"' has no exported member 'CloudFunctionsContext'.ts(2305)
```

I resolved this by changing the line to
```
import { CloudFunctionsContext } from '@google-cloud/functions-framework/build/src/functions';
```

but it suggests a couple questions:

1) Was I (and am I know) using the types as intended, and
2) Was this an intentional change (I did see some refactors mentioned in the release log), or should `CloudFunctionsContext` still be exported from the top level?

Thanks!

grant:
Hey @techieshark,
Sorry for changing the location of the interfaces. I should have mentioned the breaking change if folks are using the interfaces in the changelog.

It looks like you caught a bug!

The interfaces are here:
https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/package.json#L10

But we moved the interfaces in this file:

They should point to here:
https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/src/functions.ts

---

We should:
- Change the package.json
- Document the TS interfaces in the README

What do you think?

PR #331, opened 2021-09-22
refactor!: use a union type for SignatureType

matthewrobertson:
This commit refactors the SignatureType enum into an array of strings declared [as const](https://github.com/Microsoft/TypeScript/pull/29510). This allows the `SignatureType` type to be expressed as a union type, which works a bit better when parsing a user provided string.

This is some simple refactoring in preparation for declarative function signatures.

BREAKING CHANGE: exported SignatureType type is converted from an enum to a union type

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -36,7 +36,7 @@ import {resolve} from 'path';
 import {getUserFunction} from './loader';
 import {ErrorHandler} from './invoker';
 import {getServer} from './server';
-import {SignatureType} from './types';
+import {SignatureType, isValidSignatureType} from './types';
 
 // Supported command-line flags
 const FLAG = {
@@ -54,10 +54,6 @@ const ENV = {
   SOURCE: 'FUNCTION_SOURCE',
 };
 
-enum NodeEnv {
-  PRODUCTION = 'production',
-}
-
 const argv = minimist(process.argv, {
   string: [FLAG.PORT, FLAG.TARGET, FLAG.SIGNATURE_TYPE],
 });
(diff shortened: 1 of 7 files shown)


src/types.ts
```diff
@@ -16,8 +16,22 @@
 // executing the client function.
 export const FUNCTION_STATUS_HEADER_FIELD = 'X-Google-Status';
 
-export enum SignatureType {
```
grant, line 19: I don't agree with using string literals and arrays instead of enums. Enums provide benefits over literal values such as better typing, autocompletion, and less code here. We don't plan an arbitrary signature type until the FF contract changes.

This is also a breaking change to the API, not an internal refactor.

Could we discuss a bit more about the tradeoff in the PR / an issue?

grant (changes requested):

src/index.ts
```diff
@@ -108,7 +101,7 @@ getUserFunction(CODE_LOCATION, TARGET).then(userFunction => {
 
   SERVER.listen(PORT, () => {
     ERROR_HANDLER.register();
-    if (process.env.NODE_ENV !== NodeEnv.PRODUCTION) {
+    if (process.env.NODE_ENV !== 'production') {
```
grant, line 104: I'm not seeing TS error for misconfigured strings with all.

For example, if I type:
```ts
if (process.env.NODE_ENV !== 'prodduction') {
```

here, I won't get a TS error. I'd rather just be able to type `NodeEnv.` and see an autocompleted result of options and errors if I mistype.

src/types.ts
grant: We're fine changing to the union types, but we shouldn't completely remove the enum as it's an exported interface, unless we have strong reason to introduce a breaking change.

For example, we had a customer complaint about a similar small interface change:
https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/216

matthewrobertson: I don't think we give customer an opportunity to pass the value to use in any meaningful way (the signature type can only be provided as an env var or cli flag), so I don't think we should be too concerned about changing this. I see your point about why this is breaking change now. I will updated the conventional commit tag.

grant (approved):

PR #333, opened 2021-09-23
refactor: options parsing

matthewrobertson:
This commit refactors the logic for parsing CLI flags and env vars into a separate module and adds unit tests.

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -1,0 +1,142 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as minimist from 'minimist';
+import {resolve} from 'path';
+import {SignatureType, isValidSignatureType} from './types';
+
+/**
+ * Error thrown when an invalid option is provided.
+ */
+export class OptionsError extends Error {}
+
+/**
+ * The set of all options that can be used to configure the behaviour of
+ * the framework.
+ */
+export interface FrameworkOptions {
+  /**
+   * The port on which this server listens to all HTTP requests.
+   */
+  port: string;
+  /**
+   * The name of the function within user's node module to execute. If such a
+   * function is not defined, then falls back to 'function' name.
+   */
+  target: string;
+  /**
+   * The path to the source code file containing the client function.
+   */
+  sourceLocation: string;
+  /**
+   * The signature type of the client function.
+   */
+  signatureType: SignatureType;
+  /**
+   * Whether or not the --help CLI flag was provided.
+   */
+  printHelp: boolean;
+}
+
+/**
+ * Helper class for parsing an configurable option from provided CLI flags
+ * or environment variables.
+ */
+class ConfigurableOption<T> {
+  constructor(
+    /**
+     * The CLI flag that can be use to configure this option.
+     */
+    public readonly cliOption: string,
+    /**
+     * The name of the environment variable used to configure this option.
+     */
+    private envVar: string,
+    /**
+     * The default value used when this option is not configured via a CLI flag
+     * or environment variable.
+     */
+    private defaultValue: T,
+    /**
+     * A function used to valid the user provided value.
+     */
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    private validator: (x: any) => T = x => x as T
+  ) {}
+
+  parse(cliArgs: minimist.ParsedArgs, envVars: NodeJS.ProcessEnv): T {
+    return this.validator(
+      cliArgs[this.cliOption] || envVars[this.envVar] || this.defaultValue
+    );
+  }
+}
+
+const PortOption = new ConfigurableOption('port', 'PORT', '8080');
+const FunctionTargetOption = new ConfigurableOption(
+  'target',
+  'FUNCTION_TARGET',
+  'function'
+);
+const SourceLocationOption = new ConfigurableOption(
+  'source',
+  'FUNCTION_SOURCE',
+  '',
+  resolve
+);
+const SignatureOption = new ConfigurableOption(
+  'signature-type',
+  'FUNCTION_SIGNATURE_TYPE',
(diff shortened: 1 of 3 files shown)


src/options.ts
```diff
@@ -0,0 +1,125 @@
...
+ * or environment variables.
+ */
+class ConfigurableOption<T> {
+  constructor(
+    public readonly cliOption: string,
+    private envVar: string,
+    private defaultValue: T,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    private validator: (x: any) => T = x => x as T
+  ) {}
```
grant, line 35: nit: I think this constructor would be more readable as an object with these 4 ordered fields.

I don't think the interface it produces is easily understandable:

```ts
new ConfigurableOption(
  'target',
  'FUNCTION_TARGET',
  'function'
);
```

Suggestion:

```js
new ConfigurableOption({
  cliOption: 'target',
  envVar: 'FUNCTION_TARGET',
  defaultValue: 'function',
});
```

or skipping the `ConfigurableOption` class altogether.

src/options.ts
```diff
@@ -0,0 +1,125 @@
...
+    if (isValidSignatureType(x)) {
+      return x;
+    }
+    throw new OptionsError(
+      `Function signature type must be one of: ${SignatureType.join(', ')}.`
+    );
+  }
+);
+
+export interface FrameworkOptions {
```
grant, line 70: nit: comment missing

src/options.ts
```diff
@@ -0,0 +1,125 @@
...
+   */
+  sourceLocation: string;
+  /**
+   * The signature type of the client function.
+   */
+  signatureType: SignatureType;
+  /**
+   * Whether or not the --help CLI flag was provided.
+   */
+  printHelp: boolean;
```
grant, line 91: This Framework option seems odd. Should the interface be called `FrameworkCLIOptions` or should this boolean not be part of the main interface?

src/options.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+export const helpText = `Example usage:
```
grant, line 94: This help text doesn't seem like an option. I'm not sure how it is supposed to be used either.
Comment?

I don't know if we want to include it with this PR (we might want a more robust help text).

src/options.ts
```diff
@@ -0,0 +1,125 @@
...
+  'http' as SignatureType,
+  x => {
+    if (isValidSignatureType(x)) {
+      return x;
+    }
+    throw new OptionsError(
+      `Function signature type must be one of: ${SignatureType.join(', ')}.`
+    );
+  }
+);
```
grant, line 68: I feel like all of these options could just be put in an array/object, and not 4 constants. Then below you could just use that array directly.

src/options.ts
```diff
@@ -0,0 +1,125 @@
...
+      FunctionTargetOption.cliOption,
+      SignatureOption.cliOption,
+      SourceLocationOption.cliOption,
+    ],
+  });
+  return {
+    port: PortOption.parse(argv, envVars),
+    target: FunctionTargetOption.parse(argv, envVars),
+    sourceLocation: SourceLocationOption.parse(argv, envVars),
+    signatureType: SignatureOption.parse(argv, envVars),
```
grant, line 122: Per comment above, this seems like it could be simplified as we're just adding a list of `ConfigurableOption`. Doesn't seem to pass the squint test.

test/options.ts
```diff
@@ -0,0 +1,131 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as assert from 'assert';
+import {resolve} from 'path';
+import {parseOptions, FrameworkOptions} from '../src/options';
+
+describe('parseOptions', () => {
+  const testData = [
```
grant, line 20: Could we add the TS interface for the test data? I've found that it's very useful later on.

grant (changes requested):
Thanks. Looks good.

I think there are some parts where we can simplify the implementation and SLOCs while keeping the same behavior. I've added some comments.

src/options.ts
matthewrobertson: I agree we should probably have some more advanced help text functionality, but that is an issue for another day. 

For now, I moved this message into this module instead of leaving into index.js for the sake of cohesion. The help text is derived from cli options, so it should live in the same module. This way if someone comes along in the future and adds a new CLI flag, they are more likely to update the help text.

src/options.ts
matthewrobertson: I can remove the class and convert it to an interface in function if you prefer.

src/options.ts
matthewrobertson: I think we need to rework the `--help` stuff in a different commit. Ideally this would be handled by a more sophisticated CLI library like [yargs](https://www.npmjs.com/package/yargs), but I don't want to pull that into the functions framework because it would pull along a bunch of transitive dependencies into customers' functions.

src/options.ts
matthewrobertson: I spent an hour trying to change this, but I couldn't get it to come out any cleaner. I either ended up with a bunch more code, or losing the strong typing offered by the `FrameworkOptions` interface. Could we land this as it is and try and take another pass at cleanup in subsequent CL?

src/options.ts
matthewrobertson: see comment below about difficulty of cleaning this up.

src/options.ts
matthewrobertson: see comment below about difficulty of cleaning this up.

test/options.ts
```diff
@@ -0,0 +1,138 @@
...
+      name: 'respects all cli flags',
+      cliOpts: [
+        'bin/node',
+        '/index.js',
+        '--port',
+        '1234',
+        '--target=helloWorld',
+        '--signature-type',
+        'cloudevent',
+        '--source=/source',
```
grant, line 66: nit: we're mixing using ` ` and `=` for arguments. Can we choose one (`=` seems cleaner)? Here and elsewhere.

test/options.ts
```diff
@@ -0,0 +1,138 @@
+// Copyright 2019 Google LLC
```
grant, line 1: nit: We should be consistent throughout this PR of the header (there's 2019, 2021), unless these files are copied, in which case we don't need to touch the header.

src/options.ts
grant: OK np.

grant (approved):
Small requested changes, approved.

test/options.ts
matthewrobertson: This was intentional to make sure we get test coverage for both syntax options. Alternatively I could have split this into two tests cases for `--this case` and `--this=case`. But I thought this was a bit more terse.

PR #335, opened 2021-09-24
refactor!: move function wrapping earlier

matthewrobertson:
This commit extracts the logic for wrapping user functions into its
own module and adds unit tests. It also moves the code path that
wraps the user function earlier in the initialization logic. This
removes user function wrapping from the critical request path.


src/wrappers.ts
```diff
@@ -0,0 +1,231 @@
+// Copyright 2021 Google LLC
```
grant, line 1: Could we rename `wrappers.ts` to something more descriptive like `functionWrappers.ts`?

grant (approved):

src/wrappers.ts
matthewrobertson: sure

Issue #349, opened 2021-10-08
Add api-extractor


matthewrobertson:
We should include [api-extractor](https://api-extractor.com/) in our build pipeline to automatically extract markdown API docs from our typescript / jsdoc annotations. Checking the md files into our repo also makes it easy to see how the public API of the npm package has changed and determine whether or not a given PR is a breaking change.

PR #354, opened 2021-10-15
docs: add API extractor

grant:
Add setup for creating API docs for the Node Functions Framework. Does not attempt to annotate whole library (but just enough to make the generator not err).

Fixes https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/349

Generated API docs: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/grant_api_extractor/docs/api.md

---

Note: There aren't that many types right now because we removed a large chunk of the exported API: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/347/files#r724614498

I'm not sure though.

I think we're trying to add that back before releasing a new version of this framework.

### DIFF
--- a/docs/generated/api.d.ts
+++ b/docs/generated/api.d.ts
@@ -108,8 +108,9 @@ export declare type Context = CloudFunctionsContext | CloudEventsContext;
 
 /**
  * A data object used for legacy event functions.
+ * @public
  */
-declare interface Data {
+export declare interface Data {
     data: object;
 }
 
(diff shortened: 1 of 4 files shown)


matthewrobertson:
> There aren't that many types right now because we removed a large chunk of the exported API

Something is fishy here, but I don't think we removed the exported API. I see some issues that suggest api-extractor has trouble with re-export statements: https://github.com/microsoft/rushstack/issues/663

Maybe we need to just update [this line](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/src/index.ts#L19) in `src/index.ts`

PR #365, opened 2021-10-19
chore: bump CI to node12+

grant:
Node 10 is deprecated and unsupported. We shouldn't be running CI on that version.

This PR removes node10 and updates scripts to use the latest Node LTS version.

Related: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/364


matthewrobertson:
I don't think should do this yet. Nodejs10 still sees a lot of usage on GCF.

PR #379, opened 2021-11-05
refactor!: rename cloudevent to CloudEvent

### LINKED ISSUE #378: CloudEvent registration method should be camelCase
"Cloud Event" is two words so the "E" should be capitalized:

```typescript
import * as functions from '@google-cloud/functions-framework';

functions.cloudEvent('helloCloudEvent', (cloudEvent) => {
  console.log('Hello, CloudEvent!');
});
```

matthewrobertson:
This commit is a refactor of all our CloudEvent identifiers to be more
consistent. Depending on the context we will always prefer: `CloudEvent`,
`cloudEvent`, or `cloud_event`. We will avoid `cloudevent`

Fixes #378


anniefu (approved):

grant (changes requested):
Looks good, but I believe `cloud_events` should be `cloudevents` since CloudEvents is one word, not Cloud Events.

PR #381, opened 2021-11-05
docs: update cloudevent function docs to use declarative API

### LINKED ISSUE #357: Document Declarative Function Signatures
We should document the interface to declarative function signatures on the main `README.md`. It's not clear how to use the interface or how it interacts with non-declarative functions. There are many edge cases to consider, examples:

- What happens if there are multiple functions with the same name?
- What are valid function names? Do these align with GCF?
- What if both a declarative and non-declarative function has been specified?

Implementation: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/347

matthewrobertson:
This commit updates our README and docs/ to use declarative function signatures for CloudEvent functions.

fixes #357


README.md
```diff
@@ -228,18 +228,20 @@ It will be passed as an argument to your function when it receives a request.
...
   console.log(cloudevent.subject);
   console.log(cloudevent.id);
   console.log(cloudevent.time);
   console.log(cloudevent.datacontenttype);
-};
+});
 ```
 
-To enable CloudEvents, set the signature type to `cloudevent`. By default, the HTTP signature will be used and automatic event unmarshalling will be disabled.
+To enable the CloudEvent functions, you must list the Functions Framework as a dependency in your `package.json`.
```
anniefu, line 244: Can we add the `package.json` snippet here?
```json
{
  "dependencies": {
    "@google-cloud/functions-framework": "some version we recommend",
    }
}
```

Since people can deploy functions without testing/having `npm` locally.

anniefu (approved):

README.md
matthewrobertson: sure

grant (approved):

Issue #380, opened 2021-11-05
Spell CloudEvents Uniformly


grant:
In this repo, it appears we have multiple spellings of the term CloudEvent, in particular with TypeScript filenames:

- `CloudEvent.ts`
- `cloud_events.ts`
- `CloudEvents.ts`
- `cloudevent.ts`

Ideally, these spellings should be uniform within the repo. Having the spelling uniform across the FFs (but idiomatic to the language) would be great too.

PR #382, opened 2021-11-06
refactor!: use snake case for generated CloudEvent files

matthewrobertson:
This is consistent with other filenames in the project.


grant (approved):
LGTM.
I believe this references / fixes: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/380

Issue #388, opened 2021-11-12
Add validation for function names


matthewrobertson:
Currently the declarative registration API allows customer to use any string when registering their functions. We should constrain this to characters that can appear in a URL path.

PR #390, opened 2021-11-18
ci: add basic validation for declarative function names

grant:
Adds basic function name validation for registered functions. As almost any sequence of strings can be URL encoded into a valid URL when URL encoded, we check for valid function names using a regex.

Fixes https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/388

### DIFF
--- a/src/function_registry.ts
+++ b/src/function_registry.ts
@@ -44,9 +44,11 @@ const register = (
 };
 
 /**
- * Returns true if the function name is valid (lowercase alphanumeric, numbers, or dash characters).
- * Does not check for beginning or ending with a dash.
- * Does not check for length <= 63 characters.
+ * Returns true if the function name is valid
+ * - must contain only alphanumeric, numbers, or dash characters
+ * - must be <= 63 characters
+ * - must start with a letter
+ * - must end with a letter or number
  * @param functionName the function name
  * @returns true if the function name is valid
  */
--- a/src/function_registry.ts
+++ b/src/function_registry.ts
@@ -33,11 +33,28 @@ const register = (
   signatureType: SignatureType,
   userFunction: HandlerFunction
 ): void => {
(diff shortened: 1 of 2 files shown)


src/function_registry.ts
```diff
@@ -33,12 +33,29 @@ const register = (
...
+/**
+ * Returns true if the function name is valid (lowercase alphanumeric or dash characters).
+ * Does not check for beginning or ending with a dash.
+ * Does not check for length <= 63 characters.
+ * @param functionName the function name
+ * @returns true if the function name is valid
+ */
+export const isValidFunctionName = (functionName: string): boolean => {
+  // Validate function name with alpha characters, numbers, and dashes
+  const regex = /^[a-zA-Z-]+$/;
```
matthewrobertson, line 55: Can we also impose the restrictions that the function name
- must be <= 63 characters
- must start with a letter 
- must end with a letter or number

Something like:

```
const regex = /[A-Za-z](?:[-_A-Za-z0-9]{0,61}[A-Za-z0-9])?/;
```

grant: We can, although I don't know if we should restrict the function name to exactly what a specific vendor requires, so I didn't make the regex too tight.

WDYT?

matthewrobertson: I don't think these limitations are so onerous that they will degrade the usability for non-GCF customers.

grant: I disagree that non-Google customer would not want a function name outside of these restrictions, such as 70 characters or non-alphanumeric characters. I believe the FF shouldn't restrict its usage for the limitations of one specific platform.

However, I've implemented your suggestion as it doesn't matter too much.

matthewrobertson: Thank you!

matthewrobertson (approved):
LGTM

PR #392, opened 2021-11-20
feat: add testing helpers and instructions

### LINKED ISSUE #387: Document the unit testing story
We should write some internal guide that explains how to unit test functions. Things we should address:

- Make sure declarative function signatures provides a mechanism for customers to easily invoke functions in tests
- Writing integration tests for the express handlers with [supertest](https://www.npmjs.com/package/supertest)

matthewrobertson:
This commit adds some helpers for unit testing cloud functions along with some documentation about how to use them. Much of inspiration came from the Ruby Functions Framework: https://github.com/GoogleCloudPlatform/functions-framework-ruby/blob/main/docs/testing-functions.md

A readable preview of the docs is [here](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/f2ce0b97e2e1f188e3e892da07404127b1ecbe20/docs/testing-functions.md).

Eventually we will want to update [this guide on DevSite](https://cloud.google.com/functions/docs/testing/test-overview) to match the guidance in the docs we are adding in this PR; however, at the moment none of our public docs or samples are using the declarative function signature APIs so the helpers / instructions here are not compatible. Once the majority of the nodejs public documentation is updated to use the declarative APIs we can update the testing guide.

I also refactored some of our out tests to use the helpers.

Fixes #387

### DIFF
--- a/docs/testing-functions.md
+++ b/docs/testing-functions.md
@@ -1,0 +1,201 @@
+<!--
+# @title Testing Functions
+-->
+
+# Testing Functions
+
+This guide covers writing unit tests for functions using the Functions Framework
+for Node.js.
+
+## Overview of function testing
+
+One of the benefits of the functions-as-a-service paradigm is that functions are
+easy to test. In many cases, you can simply call a function with input, and test
+the output. You do not need to set up (or mock) an actual server.
+
+The Functions Framework provides utility methods that streamline the process of
+setting up functions and the environment for testing, constructing input
+parameters, and interpreting results. These are available in the
+`@google-cloud/functions-framework/testing` module.
+
+## Loading functions for testing
+
+The easiest way to get started unit testing Node.js Cloud Functions is to explicitly
+export the functions you wish to unit test.
+
+```js
+// hello_tests.js
+import * as functions from '@google-cloud/functions-framework';
+
+// declare a cloud function and export it so that it can be
+// imported in unit tests
+export const HelloTests = (req, res) => {
+  res.send('Hello, World!');
+};
+
+// register the HelloTests with the Functions Framework
+functions.http('HelloTests', HelloTests);
+```
+
+This is a perfectly acceptable approach that allows you to keep your application
+code decoupled from the Functions Framework, but it also has some drawbacks.
+You won't automatically benefit from the implicit type hints and autocompletion
+that are available when you pass a callback to `functions.http` directly:
+
+```js
+// hello_tests.js
+import * as functions from '@google-cloud/functions-framework';
+
+// register the HelloTests with the Functions Framework
+functions.http('HelloTests', (req, res) => {
+  // req and res are strongly typed here
+});
+```
+
+The testing module provides a `getFunction` helper method that can be used to
+access a function that was registered with the Functions Framework. To use it in
+your unit test you must first load the module that registers the function you wish
+to test.
+
+```js
+import {getFunction} from "@google-cloud/functions-framework/testing";
+
+describe("HelloTests", () => {
+  before(async () => {
+    // load the module that defines HelloTests
+    await import("./hello_tests.js");
+  });
+
+  it("is testable", () => {
+    // get the function using the name it was registered with
+    const HelloTest = getFunction("HelloTests");
+    // ...
+  });
+});
+```
+
+## Testing HTTP functions
+
+Testing an HTTP function is generally as simple as generating a request, calling
+the function, and asserting against the response.
+
+HTTP functions are passed an [express.Request](https://expressjs.com/en/api.html#req)
+and an [express.Response](https://expressjs.com/en/api.html#res) as arguments. You can
+create simple stubs to use in unit tests.
+
+```js
+import assert from "assert";
+import {getFunction} from "@google-cloud/functions-framework/testing";
+
+describe("HelloTests", () => {
+  before(async () => {
+    // load the module that defines HelloTests
+    await import("./hello_tests.js");
+  });
+
+  it("is testable", () => {
+    // get the function using the name it was registered with
+    const HelloTest = getFunction("HelloTests");
+
+    // a Request stub with a simple JSON payload
+    const req = {
+      body: { foo: "bar" },
+    };
+    // a Response stub that captures the sent response
+    let result;
+    const res = {
+      send: (x) => {
+        result = x;
+      },
+    };
+
+    // invoke the function
+    HelloTest(req, res);
+
+    // assert the response matches the expected value
+    assert.equal(result, "Hello, World!");
+  });
+});
+```
+
+## Testing CloudEvent functions
+
+Testing a CloudEvent function works similarly. The
+[JavaScript SDK for CloudEvents](https://github.com/cloudevents/sdk-javascript) provides
+APIs to create stub CloudEvent objects for use in tests.
+
+Unlike HTTP functions, event functions do not accept a response argument. Instead, you
+will need to test side effects. A common approach is to replace your function's
+dependencies with mock objects that can be used to verify its behavior. The
+[sinonjs](https://sinonjs.org/) is a standalone library for creating mocks that work with
+any Javascript testing framework:
+
+```js
+import assert from "assert";
+import sinon from "sinon";
+import {CloudEvent} from "cloudevents";
+import {getFunction} from "@google-cloud/functions-framework/testing";
+
+import {MyDependency} from "./my_dependency.js";
+
+describe("HelloCloudEvent", () => {
+  before(async () => {
+    // load the module that defines HelloCloudEvent
+    await import("./hello_cloud_event.js");
+  });
+
+  const sandbox = sinon.createSandbox();
+
+  beforeEach(() => {
+    sandbox.spy(MyDependency);
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it("uses MyDependency", () => {
+    const HelloCloudEvent = getFunction("HelloCloudEvent");
+    HelloCloudEvent(new CloudEvent({
+      type: 'com.google.cloud.functions.test',
+      source: 'https://github.com/GoogleCloudPlatform/functions-framework-nodejs',
+    }));
+    // assert that the cloud function invoked `MyDependency.someMethod()`
+    assert(MyDependency.someMethod.calledOnce);
+  });
+});
+```
+
+## Integration testing with SuperTest
+
+The `testing` module also includes utilities that help you write high-level, integration
+tests to verify the behavior of the Functions Framework HTTP server that invokes your function
+to respond to requests. The [SuperTest](https://github.com/visionmedia/supertest) library
+provides a developer friendly API for writing HTTP integration tests in javascript. The
+`testing` module includes a `getTestServer` helper to help you test your functions using
+SuperTest.
+
+```js
+import supertest from 'supertest';
+import {getTestServer} from '@google-cloud/functions-framework/testing';
+
+describe("HelloTests", function () {
+  before(async () => {
+    // load the module that defines HelloTests
+    await import("./hello_tests.js");
+  });
+
+  it("uses works with SuperTest", async () => {
+    // call getTestServer with the name of function you wish to test
+    const server = getTestServer("HelloTests");
+
+    // invoke HelloTests with SuperTest
+    await supertest(server)
+      .post("/")
+      .send({ some: "payload" })
+      .set("Content-Type", "application/json")
+      .expect("Hello, World!")
+      .expect(200);
+  });
+});
+```
--- a/src/testing.ts
+++ b/src/testing.ts
@@ -1,0 +1,55 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This module provides a set of utility functions that are useful for unit testing Cloud Functions.
+import {Server} from 'http';
+
+import {HandlerFunction} from '.';
+import {getRegisteredFunction} from './function_registry';
+import {getServer} from './server';
+
+/**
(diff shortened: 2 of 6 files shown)


docs/testing-functions.md
```diff
@@ -0,0 +1,200 @@
...
+```
+
+## Testing HTTP functions
+
+Testing an HTTP function is generally as simple as generating a request, calling
+the function, and asserting against the response.
+
+HTTP functions are passed an [express.Request](https://expressjs.com/en/api.html#req) 
+and an [express.Response](https://expressjs.com/en/api.html#res) as arguments. It is
+not difficult to create simple stubs to use in unit tests.
```
anniefu, line 85: [nit] maybe just "You can create simple stubs to use in unit tests"

test/integration/cloud_event.ts
```diff
@@ -50,35 +50,25 @@ describe('CloudEvent Function', () => {
...
-      body: TEST_CLOUD_EVENT,
-      expectedCloudEvent: {
-        ...TEST_CLOUD_EVENT,
-        traceparent: '00-65088630f09e0a5359677a7429456db7-97f23477fb2bf5ec-01',
-      },
+      body: testCloudEvent(),
+      expectedCloudEvent: testCloudEvent({
+        // TODO: the traceparent extension only seems to get set for binary
+        // cloudevent requests.
+        // traceparent: '00-65088630f09e0a5359677a7429456db7-97f23477fb2bf5ec-01',
```
matthewrobertson, line 65: @grant this seems to be a bug with the CE traceparent stuff. The previous test wasn't actually validating this correctly so we didn't notice.

anniefu (approved):
Neat! I prefer the testing package approach to exposing getters on the FF registration package. 

Next up, we generate the test stubs for you 😃

grant (changes requested):
Thanks for the PR! I've given a review.

Given there's a lot of content: docs, samples, and implementation, we might want to discuss this PR a bit more. Are we going to provide these helpers uniformly across languages? The changes here may also impact other teams.

For some context, the origin of the `docs/` folder was to cover topics that aren't covered DevSite and provide a lightweight guide for users using this framework. Some of the frameworks provide guidance within the repo, with guidance also being outside the repo.

src/testing.ts
```diff
@@ -0,0 +1,128 @@
...
+  data: {
+    test: 'payload',
+  },
+};
+
+/**
+ * The CloudEvent fields that should be sent as HTTP headers in binary
+ * CloudEvent requests.
+ */
+const BINARY_CLOUD_EVENT_HEADER_FIELDS: (keyof CloudEventsContext)[] = [
```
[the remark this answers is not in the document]
matthewrobertson: As per your comment below, we can ditch this and the `toBinaryCloudEvent` util method below

src/testing.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  if (cloudEvent.traceparent) {
+    headers['traceparent'] = cloudEvent.traceparent;
+  }
+
+  return {
+    headers,
+    body: cloudEvent.data,
+  };
+};
```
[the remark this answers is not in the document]
matthewrobertson: Sure let's just leave this out for now. We might be able to do something different depending on our decision about he CESDK / google-events libs.

test/integration/cloud_event.ts
matthewrobertson: this is fixed now

src/testing.ts
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Create a stub CloudEvent object that can be used to invoke CloudEvent functions
+ * @param properties a set of properties to overried in the CloudEvent
+ * @returns an object that can be used to invoke a CloudEvent function
+ *
+ * @beta
+ */
+export const testCloudEvent = (
```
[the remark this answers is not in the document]
matthewrobertson: I think it nice for us to provide an API creating stubs of the objects we pass to users functions. This is something we use heavily in our own tests so I suspect developers will get some value out of it as well.

test/integration/legacy_event.ts
```diff
@@ -13,22 +13,17 @@
...
-  subject: 'test-subject',
-  id: 'test-1234-1234',
-  time: '2020-05-13T01:23:45Z',
-  traceparent: '00-65088630f09e0a5359677a7429456db7-97f23477fb2bf5ec-01',
-  datacontenttype: 'application/json',
-  data: {
-    some: 'payload',
-  },
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const removeData = <T extends {[key: string]: any}>(foo: T): T => {
```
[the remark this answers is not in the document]
matthewrobertson: I removed this. It isn't really in scope of the change I am trying to make here.

docs/testing-functions.md
```diff
@@ -0,0 +1,198 @@
...
+parameters, and interpreting results. These are available in the
+`@google-cloud/functions-framework/testing` module.
+
+## Loading functions for testing
+
+The easiest way to get started unit testing Node.js cloud functions is to explicitly
+export the functions you wish to unit test.
+
+```javascript
+const functions = require('@google-cloud/functions-framework');
```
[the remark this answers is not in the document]
matthewrobertson: I updated all the samples to use ESM.

src/testing.ts
```diff
@@ -0,0 +1,128 @@
...
+ * @param functionName the name of the function to get
+ * @returns a function that was registered with the Functions Framework
+ *
+ * @beta
+ */
+export const getFunction = (
+  functionName: string
+): HandlerFunction | undefined => {
+  return getRegisteredFunction(functionName)?.userFunction;
+};
```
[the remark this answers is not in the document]
matthewrobertson: I think this is a crucial part of the `testing` API. `getRegisteredFunction` is not part of our public API. It is a private implementation detail that is subject to change. For example right now it returns the function and signature type so we can wrap it later, but we might change it to wrap the function immediately upon registration and drop the signature type later if it make more sense.

docs/testing-functions.md
```diff
@@ -0,0 +1,198 @@
...
+
+describe("HelloCloudEvent", () => {
+  before(async () => {
+    // load the module that defines HelloCloudEvent
+    await import("./hello_cloud_event.js");
+  });
+
+  const sandbox = sinon.createSandbox();
+
+  beforeEach(function () {
```
grant, line 149: nit: Use arrow functions consistently throughout this project / file. `beforeEach(() => {`

Here and elsewhere.

src/testing.ts
```diff
@@ -0,0 +1,83 @@
...
```
grant, line 48: Discussion: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/392#discussion_r755617031

I believe it would be reasonable to change the location of the export. Specifically, export `getFunction` with the other function registry exports here:
https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/src/index.ts

That's my suggestion.

---

I still believe this useful function is not necessarily tied to testing and am suggesting we simply move the export location. There is no testing logic in this function and the same `@beta` tag can still be provided there.

We shouldn't mix import locations for testing and registry functions.

src/testing.ts
```diff
@@ -0,0 +1,83 @@
...
+ * Create a stub CloudEvent object that can be used to invoke CloudEvent functions
+ * @param properties a set of properties to override in the CloudEvent
+ * @returns an object that can be used to invoke a CloudEvent function
+ *
+ * @beta
+ */
+export const testCloudEvent = (
+  properties: Partial<CloudEventsContext> = {}
+): CloudEventsContext => {
+  return {...TEST_CLOUD_EVENT, ...properties};
```
grant, line 82: I'd `Object.freeze` this object and remove all logic that assumes this object can be mutable (see below).

src/testing.ts
```diff
@@ -0,0 +1,83 @@
...
+import {Server} from 'http';
+
+import {CloudEventsContext, HandlerFunction} from '.';
+import {getRegisteredFunction} from './function_registry';
+import {getServer} from './server';
+
+/**
+ * The default CloudEvent returned by testCloudEvent
+ */
+const TEST_CLOUD_EVENT: CloudEventsContext = {
```
grant, line 25: It looks like we remove `subject: 'test-subject',` here. Can we add it back? That's a valid CE property that we should test.

test/integration/cloud_event.ts
```diff
@@ -53,33 +47,33 @@ describe('CloudEvent Function', () => {
...
-        'ce-id': TEST_CLOUD_EVENT.id,
-        'ce-time': TEST_CLOUD_EVENT.time,
-        'ce-datacontenttype': TEST_CLOUD_EVENT.datacontenttype,
+        'ce-specversion': testCloudEvent().specversion,
+        'ce-type': testCloudEvent().type,
+        'ce-source': testCloudEvent().source,
+        'ce-subject': 'test-subject',
+        'ce-id': testCloudEvent().id,
+        'ce-time': testCloudEvent().time,
+        'ce-datacontenttype': testCloudEvent().datacontenttype,
```
grant, line 71: I believe it's pretty odd to call a function 7 times to get the properties of the returned object. We're using the same CloudEvent object.

In this instance, we're getting 7 different objects, then just getting one property from them.

Can we keep using the static CloudEvent, especially if it is frozen?

grant:
Thanks, reviewed! Getting there.

Do you have any thoughts on the comment from before?

> Given there's a lot of content: docs, samples, and implementation, we might want to discuss this PR a bit more. Are we going to provide these helpers uniformly across languages? The changes here may also impact other teams.
> 
> For some context, the origin of the `docs/` folder was to cover topics that aren't covered DevSite and provide a lightweight guide for users using this framework. Some of the frameworks provide guidance within the repo, with guidance also being outside the repo.

We typically defer heavier documentation to DevSite docs as there can be a mix of guidance.

grant (changes requested):
Thanks, reviewed! Getting there.

Do you have any thoughts on the comment from before?

> Given there's a lot of content: docs, samples, and implementation, we might want to discuss this PR a bit more. Are we going to provide these helpers uniformly across languages? The changes here may also impact other teams.
> 
> For some context, the origin of the `docs/` folder was to cover topics that aren't covered DevSite and provide a lightweight guide for users using this framework. Some of the frameworks provide guidance within the repo, with guidance also being outside the repo.

We typically defer heavier documentation to DevSite docs as there can be a mix of guidance.

grant:
It would help to have more details as to how this interface is expected to be used here. This utility is useful.

This PR is large, substantial, and external-facing. It includes both implementation and documentation. Here's an example PR description that would help:

---

feat: add testing helpers and instructions

This PR blocks https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/2428

Feat:

- Adds a `@google-cloud/functions-framework/testing` package
  - Includes X, Y, Z interfaces
- This testing utility is one-off for the Node FF

Docs:

- Adds a guide on testing functions
  - sample for loading, registering, and running tests using assert
  - testing HTTP functions
  - testing CloudEvent functions

This content should not go in our docs because of X, Y:

- https://cloud.google.com/functions/docs/testing/test-http
- https://cloud.google.com/functions/docs/testing/test-event

I plan to write this testing guide for Node.js only. We'll follow-up with each language individually.

src/testing.ts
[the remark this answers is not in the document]
matthewrobertson: OK I removed `testCloudEvent` and updated the docs to point developers to the [JavaScript SDK for CloudEvents](https://github.com/cloudevents/sdk-javascript) for this type of functionality.

docs/testing-functions.md
```diff
@@ -0,0 +1,198 @@
...
+    sandbox.spy(MyDependency);
+  });
+
+  afterEach(function () {
+    sandbox.restore();
+  });
+
+  it("uses MyDependency", () => {
+    const HelloCloudEvent = getFunction("HelloCloudEvent");
+    HelloCloudEvent(testCloudEvent());
```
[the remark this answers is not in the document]
matthewrobertson: updated with a suggestion to use the [JavaScript SDK for CloudEvents](https://github.com/cloudevents/sdk-javascript)

src/testing.ts
matthewrobertson: I removed this as per suggestions above

src/testing.ts
matthewrobertson: I removed this as per discussion above.

test/integration/cloud_event.ts
matthewrobertson: reverted this as I removed `testCloudEvent`

matthewrobertson:
> We typically defer heavier documentation to DevSite docs as there can be a mix of guidance.

@grant currently none of our Node.js docs on DevSite are using the declarative function signature APIs, so this guide and the `testing` module are incompatible with those docs / samples. Once we update all (most?) of the other docs on DevSet to use declarative I think we should update [this guide](https://cloud.google.com/functions/docs/testing/test-overview) to align it with the docs and helpers I am adding here. The plan for that is still TBD so I think checking some docs in here is a good interim solution. 

Also FWIW the [dotnet](https://github.com/GoogleCloudPlatform/functions-framework-dotnet/blob/main/docs/testing.md) and [ruby](https://github.com/GoogleCloudPlatform/functions-framework-ruby/blob/main/docs/testing-functions.md) repos already include similar markdown docs that cover testing functions.

src/testing.ts
matthewrobertson: My strong preference is to not expose this function from the primary API in `index.ts`. My primary concern is that I expect some flux in this API as we add more functionality to the framework (e.g. lifecycle events). Exporting from here sends a strong signal to developers that it's only purpose is for testing and it is not to be used in production.

grant: This is fine. I don't agree that the location of the export will change developer behavior (see recent internal code location changes for example), but the difference is insignificant.

test/integration/cloud_event.ts
```diff
@@ -71,7 +79,7 @@ describe('CloudEvent Function', () => {
         'ce-specversion': TEST_CLOUD_EVENT.specversion,
         'ce-type': TEST_CLOUD_EVENT.type,
         'ce-source': TEST_CLOUD_EVENT.source,
-        'ce-subject': TEST_CLOUD_EVENT.subject,
+        'ce-subject': 'test-subject',
```
grant, line 82: Sorry if I commented this before. I don't think the constant should be substituted with a string literal without a comment.

I think we should keep:
`TEST_CLOUD_EVENT.subject,`
or add a comment.

grant (approved):
Approved. Nothing blocking.

I think we could still use a more detailed PR description. I'm not sure if we will or want to keep docs around this topic. It doesn't seem like we're aiming to do this across FFs.

This PR is urgent, so lgtm.

Issue #394, opened 2021-11-23
CloudEvent traceparent extension not set for structured content requests


matthewrobertson:
When a `traceparent` header is sent with a structured CE request, it is not properly set as an extension attribute of the CE we pass to the customer's function.

We have [a unit test to validate this use case](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/test/integration/cloud_event.ts#L57-L66). It only passes because the `traceparent` field is already set to the expected value in the [TEST_CLOUD_EVENT](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/test/integration/cloud_event.ts#L28). We should probably update all the traceparent tests to use a different header value than we use in the `TEST_CLOUD_EVENT`.

grant:
I wouldn't consider this a bug, as we're supporting tracing for normal binary CloudEvents that we use in Eventarc. Adding support for structured CEs and a test sounds reasonable, although it won't impact GCP developers.

matthewrobertson:
The test already exists: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/test/integration/cloud_event.ts#L57-L66

And it passes even though the functionality isn't implemented. We should either fix this or delete the test.

PR #395, opened 2021-11-23
refactor: move ce extensions to separate test case. support traceparent

grant:
- Moves extensions parsing to `parseCloudEventRequest`
- Creates separate tc for extensions
- Tests binary / structured CE with `traceparent` header.

Fixes: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/394

Ideally, we don't need to manually put CloudEvent logic in the FF.


matthewrobertson (approved):
LGTM!

grant:
grazie mille

PR #400, opened 2021-12-02
Revert "Revert "refactor: use cesdk""

grant:
Reverts GoogleCloudPlatform/functions-framework-nodejs#399 (roll forward).

See PR description for details: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/391#issue-1057979568

WANT_LGTM=@matthewrobertson

PR #401, opened 2021-12-02
ci: fix docs ci

grant:
It looks like the doc CI check would not produce the right exit code. This should correctly report bad doc errors (see below).

- Fixes docs CI error code
- Runs docs CI


grant:
Ironically, docs CI failed for a different PR appropriately. 

https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/400

I suspect this CI must not fail for certain changes. Anyways, I believe this CI change is more stable.

matthewrobertson (approved):
LGTM

Issue #322, opened 2021-09-15
Errors in the function should result in non-200 response code


bvoss:
We use the function-framework in CloudRun containers for Pub/Sub event handling.

Sadly in this case Pub/Sub dose not respect the 'X-Google-Status' header. And as the http response code is 200 in case of exceptions in the user function Pub/Sub acknowledges the message.

So in addition to setting the response header 'X-Google-Status'  to 'error' the http response code should be 500.
Or it should be at least be possible to configure this behavior

grant:
Hi @bvoss,

I was able to look at this issue and it looks like we're inconsistent with our HTTP response codes.

We also lack tests here.

I've opened a PR that should fix this issue.

If you have additional tests / repro steps, feel free to add them.

### My notes

`index.js`

```js
const functions = require('@google-cloud/functions-framework');

functions.http('myfunc', (req, res) => {
  throw new Error('test');
});
```

`package.json`

```json
{
  "main": "index.js",
  "scripts": {
    "start": "npx @google-cloud/functions-framework --target myfunc"
  },
  "dependencies": {
    "@google-cloud/functions-framework": "^2.1.0"
  }
}
```

Test:

`curl -XPOST http://localhost:8080/ -I`

### Expected

```
> curl -XPOST http://localhost:8080/ -I
HTTP/1.1 500 Internal Server Error
X-Google-Status: crash
Content-Type: text/html; charset=utf-8
Content-Length: 8
ETag: W/"8-OhraUVTQ47wG45qqkAQ/rn0lV6g"
Date: Fri, 10 Dec 2021 23:58:37 GMT
Connection: keep-alive
Keep-Alive: timeout=5
```

### Actual

```
> curl -XPOST http://localhost:8080/ -I
HTTP/1.1 200 OK
X-Google-Status: crash
Content-Type: text/html; charset=utf-8
Content-Length: 8
ETag: W/"8-OhraUVTQ47wG45qqkAQ/rn0lV6g"
Date: Sat, 11 Dec 2021 00:00:21 GMT
Connection: keep-alive
Keep-Alive: timeout=5
```

grant:
I was looking over our GitHub issues and saw this one from Sep 15.

@anniefu I don't know if you were looking at this or had a suggested behavior.

anniefu:
#427 is in and should fix this.

PR #405, opened 2021-12-11
fix!: fix crash report for cloudevent functions

grant:
Fixes HTTP response code for functions that throw an error within the function.

- Fixes function crash HTTP header
- Adds new test suite for the FF logger

It looks like we were inconsistent with sending a crash response with logic in `src/function_wrappers.ts`.

## Example

Here's an example for how to send and inspect a HTTP response:

```
curl -XPOST http://localhost:8080/ -I
HTTP/1.1 500 Internal Server Error
X-Google-Status: crash
Content-Type: text/html; charset=utf-8
Content-Length: 8
ETag: W/"8-OhraUVTQ47wG45qqkAQ/rn0lV6g"
Date: Fri, 10 Dec 2021 23:58:37 GMT
Connection: keep-alive
Keep-Alive: timeout=5
```

Fixes https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/322

### DIFF
--- a/src/function_wrappers.ts
+++ b/src/function_wrappers.ts
@@ -48,9 +48,10 @@ const getOnDoneCallback = (res: Response): OnDoneCallback => {
     } else {
       res.locals.functionExecutionFinished = true;
       if (err) {
-        console.error(err.stack);
+        sendCrashResponse({err, res});
+      } else {
+        sendResponse(result, err, res);
       }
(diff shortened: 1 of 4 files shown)


grant:
Adding Annie because @anniefu is looking over HTTP response codes for the FFs.

grant:
Per chat, closing for now.

Issue #254, opened 2021-02-05
Best way to deploy TypeScript function to Cloud Function?


dmiranda2791:
I am using the TypeScript template to write the function and I faced some issues trying to deploy it using the gcloud functions deploy command.

I tried: 

* The intuitive commando: 

```
gcloud functions deploy pocReadFromBigQueryWriteToFirestore \
--runtime nodejs10 --trigger-http --allow-unauthenticated
```

Got the error 

```
Deploying function (may take a while - up to 2 minutes)...failed.                                        
ERROR: (gcloud.functions.deploy) OperationError: code=3, message=Build failed: index.js does not exist; Error ID: e1b2ac36
```

* Then, passing a source argument pointing to the build/src directory. 
```
gcloud functions deploy pocReadFromBigQueryWriteToFirestore \
--source=build/src \
--runtime nodejs12 --trigger-http --allow-unauthenticated
```
Got this error in the cli
```
Deploying function (may take a while - up to 2 minutes)...failed.                                        
ERROR: (gcloud.functions.deploy) OperationError: code=3, message=Function failed on loading user code. This is likely due to a bug in the user code. Error message: Error: please examine your function logs to see the error cause: https://cloud.google.com/functions/docs/monitoring/logging#viewing_logs. Additional troubleshooting documentation can be found at https://cloud.google.com/functions/docs/troubleshooting#logging. Please visit https://cloud.google.com/functions/docs/troubleshooting for in-depth troubleshooting documentation.
```
And this one on the functions logs:

```
2021-02-04T17:47:55.029ZpocReadFromBigQueryWriteToFirestore Provided module can't be loaded.
Default
2021-02-04T17:47:55.029ZpocReadFromBigQueryWriteToFirestore Did you list all required modules in the package.json dependencies?
```

Next, I tried copying the package.json into the build/src folder and rerun the same command as above.

```
Deploying function (may take a while - up to 2 minutes)...failed.                                        
ERROR: (gcloud.functions.deploy) OperationError: code=3, message=Build failed: > protobufjs@6.10.2 postinstall /workspace/node_modules/protobufjs
> node scripts/postinstall


> cloud-function-poc@1.0.0 prepare /workspace
> yarn run compile

yarn run v1.22.4
$ tsc
/bin/sh: 1: tsc: not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
[...]

PR #417, opened 2022-01-12
docs: add instructions for deploying with typescript codebase

johannesgiani:
Update typescript readme as proposed in [issue](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/254)

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -54,3 +54,22 @@ the Functions Framework.
     Function: helloWorld
     URL: http://localhost:8080/
     ```
+
+## Deploying with gcloud CLI
+
+1. Adjust `main` field in `package.json` to point to the compiled javascript source. `gcloud` will inspect the `main` field to detect sources:
+    
+    ```js
+      "main": "build/src/index.js",
+      ...
+    ```
+
+2. Remove `prepare` script in `package.json` created by [gts](https://github.com/google/gts). Because this script requires typescript to be installed it will fail when deploying.
+
+3. Deploy:
+
+    ```sh
+    gcloud functions deploy helloWorld \
+    --runtime nodejsXX \ # e.g. nodejs16
+    --trigger-http
+    ```
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -57,19 +57,19 @@ the Functions Framework.
 
 ## Deploying with gcloud CLI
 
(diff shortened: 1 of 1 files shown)


docs/typescript.md
```diff
@@ -54,3 +54,22 @@ the Functions Framework.
...
+      ...
+    ```
+
+2. Remove `prepare` script in `package.json` created by [gts](https://github.com/google/gts). Because this script requires typescript to be installed it will fail when deploying.
+
+3. Deploy:
+
+    ```sh
+    gcloud functions deploy helloWorld \
+    --runtime nodejsXX \ # e.g. nodejs16
```
grant, line 73: Can you hardcode nodejs16 here and remove the comment? This way folks can copy paste.

docs/typescript.md
```diff
@@ -54,3 +54,22 @@ the Functions Framework.
...
+## Deploying with gcloud CLI
+
+1. Adjust `main` field in `package.json` to point to the compiled javascript source. `gcloud` will inspect the `main` field to detect sources:
+    
+    ```js
+      "main": "build/src/index.js",
+      ...
+    ```
+
+2. Remove `prepare` script in `package.json` created by [gts](https://github.com/google/gts). Because this script requires typescript to be installed it will fail when deploying.
```
grant, line 67: Use `1.` everywhere instead of 2 and 3.

docs/typescript.md
```diff
@@ -54,3 +54,22 @@ the Functions Framework.
     Function: helloWorld
     URL: http://localhost:8080/
     ```
+
+## Deploying with gcloud CLI
+
+1. Adjust `main` field in `package.json` to point to the compiled javascript source. `gcloud` will inspect the `main` field to detect sources:
```
grant, line 60: Remove the sentence about gcloud to simplify. That sentence isn't technically correct. There is no inspection from the CLI.

docs/typescript.md
grant, line 67: Rather than:

Because this script...

Say:

This is because the `prepare` script requires typescript to be installed and will cause the function to fail to deploy if not removed.

grant (changes requested):
Thanks! I made a few requests before merging.

docs/typescript.md
johannesgiani: Thanks for commenting :)

grant (approved):
Thank you for your contribution. LGTM. Approved.

PR #408, opened 2021-12-16
Correction to execution command

gg-ayushka:
The command
```sh
node --inspect node_modules/@google-cloud/functions-framework --target=helloWorld
```
didn't work for me. Not sure if this is because of changes, npm updates or the doc was wrong in the first place.

The working command is documented in this change.

PR #420, opened 2022-01-27
fix: patching debugging docs

cstanger:
After #377 adopting the entrypoint to ./build/src/main.js, the Node Inspector does not work as expected. 

By pointing on the symlinked executable in node_modules/.bin/functions-framework (which is reflecting the right entry point) someone can use the functions framework in conjunction with the node inspector as expected.


cstanger:
Also resolves #408

grant (approved):
Verified. Thanks!

PR #427, opened 2022-02-11
fix!: send correct response code on exception

matthewrobertson:
This commit updates the http response we send when the customer's function throws to:

1. always send a 500 HTTP status
2. strip the error message from the response body when running in production

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -27,12 +27,14 @@ export function sendCrashResponse({
   res,
   callback,
   silent = false,
+  statusHeader = 'crash',
 }: {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   err: Error | any;
   res: express.Response | null;
   callback?: Function;
   silent?: boolean;
+  statusHeader?: string;
 }) {
   if (!silent) {
     console.error(err.stack || err);
@@ -43,8 +45,14 @@ export function sendCrashResponse({
   // right before sending the response, to make sure that no concurrent
   // execution sends the response between the check and 'send' call below.
   if (res && !res.headersSent) {
-    res.set(FUNCTION_STATUS_HEADER_FIELD, 'crash');
-    res.send((err.message || err) + '');
+    res.set(FUNCTION_STATUS_HEADER_FIELD, statusHeader);
+
+    if (process.env.NODE_ENV !== 'production') {
+      res.status(500);
+      res.send((err.message || err) + '');
+    } else {
+      res.sendStatus(500);
+    }
   }
   if (callback) {
     callback();
--- a/test/integration/cloud_event.ts
+++ b/test/integration/cloud_event.ts
@@ -50,11 +50,13 @@ describe('CloudEvent Function', () => {
     clock = sinon.useFakeTimers();
     // Prevent log spew from the PubSub emulator request.
     sinon.stub(console, 'warn');
+    sinon.stub(console, 'error');
   });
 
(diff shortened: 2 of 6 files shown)


src/logger.ts
```diff
@@ -43,8 +45,11 @@ export function sendCrashResponse({
   // right before sending the response, to make sure that no concurrent
   // execution sends the response between the check and 'send' call below.
   if (res && !res.headersSent) {
-    res.set(FUNCTION_STATUS_HEADER_FIELD, 'crash');
-    res.send((err.message || err) + '');
+    res.set(FUNCTION_STATUS_HEADER_FIELD, statusHeader);
+    res.status(500);
+    if (process.env.NODE_ENV !== 'production') {
+      res.send((err.message || err) + '');
```
anniefu, line 51: Can we log the error message for production so the end user can still find the info?

May also be worth including a descriptive response body message that's generic like "error encountered during function execution, see logs for more details."

test/integration/cloud_event.ts
```diff
@@ -302,4 +304,20 @@ describe('CloudEvent Function', () => {
...
+
+    // invoke the function with a CloudEvent with a string payload
+    const server = getTestServer('testTypedCloudEvent');
+    await supertest(server)
+      .post('/')
+      .send(TEST_CLOUD_EVENT)
+      .expect(res =>
+        assert.deepStrictEqual(res.headers['x-google-status'], 'error')
+      )
+      .expect(500);
```
anniefu, line 321: Would be good to check that the exception message doesn't appear in the response body.

test/integration/legacy_event.ts
```diff
@@ -188,4 +198,24 @@ describe('Event Function', () => {
...
+        timestamp: 'testTimestamp',
+        eventType: 'testEventType',
+        resource: 'testResource',
+        data: {some: 'payload'},
+      })
+      .set({'Content-Type': 'application/json'})
+      .expect(res =>
+        assert.deepStrictEqual(res.headers['x-google-status'], 'error')
+      )
+      .expect(500);
```
anniefu, line 219: Would be good to check that the exception message doesn't appear in the response body.

src/logger.ts
matthewrobertson: stack trace is still logged (as it was before) up on line 40.

test/integration/legacy_event.ts
matthewrobertson: done

anniefu (approved):

src/logger.ts
```diff
@@ -43,8 +45,11 @@ export function sendCrashResponse({
   // right before sending the response, to make sure that no concurrent
   // execution sends the response between the check and 'send' call below.
   if (res && !res.headersSent) {
-    res.set(FUNCTION_STATUS_HEADER_FIELD, 'crash');
-    res.send((err.message || err) + '');
+    res.set(FUNCTION_STATUS_HEADER_FIELD, statusHeader);
+    res.status(500);
+    if (process.env.NODE_ENV !== 'production') {
```
grant, line 50: In production environments, does this function code path send a response? Not sure why we need this if and what happens in production.

grant (approved):

src/logger.ts
matthewrobertson: Yeah good call. I will update.

matthewrobertson: Fixed

PR #428, opened 2022-02-12
chore: update cloudevents package

### LINKED ISSUE #425: Clarification on using CloudEventsContext / Context type
I have a cloud function (triggered by PubSub) I've written in Typescript and transpiled with babel which was using the `CloudEventsContext` type [exported in version `1.9.0`](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/v1.9.0/src/functions.ts#L86). When I tried to upgrade to version `2.1.1`, I now get the following errors:
```
node_modules/cloudevents/dist/index.d.ts(7,60): error TS2748: Cannot access ambient const enums when the '--isolatedModules' flag is provided.
src/index.ts(1,15): error TS2305: Module '"@google-cloud/functions-framework"' has no exported member 'CloudEventsContext'.
```

It looks like I should now be using the [`Context`](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/v2.1.1/src/functions.ts#L139) type, but them I'm running into the same `Cannot access ambient const enums` error with the `--isolatedModules` flag, which looks like it's related to this issue with cloudevents itself: https://github.com/cloudevents/sdk-javascript/issues/456.

Can anyone provide any guidance on how I should type my cloud functions with CloudEvent context going forward? Is this something that needs to be addressed in the cloudevents repo?

Thank you!

grant:
Updates cloudevents package to fix `ambient const enums` error.

Fixes #425
Lockfile update should be fixed with: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/426

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@google-cloud/functions-framework",
-  "version": "2.1.1",
+  "version": "3.0.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@google-cloud/functions-framework",
-      "version": "2.1.1",
+      "version": "3.0.0",
       "license": "Apache-2.0",
       "dependencies": {
         "body-parser": "^1.18.3",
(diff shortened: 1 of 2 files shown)


anniefu (approved):

grant:
This should be a small change, merging.

Issue #15, opened 2019-04-10
Local debug


bluelightning2k:
Hi, the readme doesn't mention anything about local debugging.

Is breakpoint debugging supported?

This would be a necessary feature for us to begin using Functions Framework. Is this supported, and can you consider updating the readme to address offline debugging (even if it's just to state that it's not supported, or to suggest a workaround).

Thanks!

stew-r:
Hi bluelightning2k -- try the following:

1. install the module (`npm install @google-cloud/functions-framework`) to `node_modules`
2. run your function using `node --inspect`, e.g., `node --inspect node_modules/@google-cloud/functions-framework <FLAGS>`

Where `<FLAGS>` correspond to Functions Framework flags (e.g., `function-target`) as documented on the README.

That should produce something like:

```
node --inspect node_modules/@google-cloud/functions-framework --function-target helloWorld
Debugger listening on ws://127.0.0.1:9229/c3574d99-20f7-42e8-b34c-dd9ca27f4876
For help see https://nodejs.org/en/docs/inspector
Serving function...
Function: helloWorld
URL: http://localhost:8080/
```

stew-r:
I also tried this out using `npx`: `npx --node-arg=--inspect @google-cloud/functions-framework --function-target=helloWorld`

As a non-Node.js developer, the `npx` tooling is pretty slick. You should also be able to pass arguments to Node using `NODE_OPTIONS`.

stew-r:
Closing this. Please re-open if this doesn't work. I'd also love to get confirmation that this works for your purposes.

bluelightning2k:
Hi @stew-r ,
This worked perfectly!

Specifically: your first method worked perfectly, even in a complex setup.
We are using WebPack to do a reasonably complicated build, and we setup VSCode to do the build, run/restart the function, and attach the debugger - all by pressing F5.

The NPX method didn't seem to support reloading in this configuration. So if you add one to the documentation (highly recommended) I'd choose the first method.


Thank you!

stew-r:
> Specifically: your first method worked perfectly, even in a complex setup.
We are using WebPack to do a reasonably complicated build, and we setup VSCode to do the build, run/restart the function, and attach the debugger - all by pressing F5.

That's awesome! I'm glad you were able to make this work so seamlessly. Do you think there would be any value in maybe writing a small blog/Medium post about your setup? I'm sure other developers could benefit from your learnings.

[...]

PR #429, opened 2022-02-12
docs: improve docs using npx for debugging

grant:
Uses `npx` for the debugging docs. Might help with windows users. Using the `.bin` folder directly seem finicky.

See: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/issues/15#issuecomment-1036834366

Tested locally with the [debugging docs](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/docs/debugging.md):

```
npx --node-options=--inspect @google-cloud/functions-framework --target=helloWorld
Debugger listening on ws://127.0.0.1:9229/0f8f426d-8456-453d-887b-e34612cc872f
For help, see: https://nodejs.org/en/docs/inspector
Serving function...
Function: helloWorld
Signature type: http
URL: http://localhost:8080/
```

### DIFF
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -19,7 +19,7 @@ npm install @google-cloud/functions-framework
 3. Run `node`, enable the inspector and run the Functions Framework:
 
 ```sh
-node --inspect node_modules/.bin/functions-framework --target=helloWorld
+npx --node-options=--inspect @google-cloud/functions-framework --target=helloWorld
 ...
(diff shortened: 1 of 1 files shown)


grant:
@matthewrobertson PTAL?

This should be a very small docs change.

grant:
> For VS Code the npx command does not trigger the IDE Debugging mode. Therefore, I would suggest keeping it with the "node" command

Sounds good. If folks have issues still, they can create a bug. Closing.

PR #438, opened 2022-03-31
Revert "Disable Express eTag response header for consistent FF behaviour across FF runtimes "

grant:
Reverts GoogleCloudPlatform/functions-framework-nodejs#433 due to failed lint check.


anniefu (approved):
Would add in the PR description a short blurb about why we are reverting.

PR #452, opened 2022-04-05
chore: update renovate match rules

grant:
We have a few too many PRs that are minor/patch versions without significant updates.

This PR adds a schema to the renovate config and sets the package rules to group minor/patch PRs.

Automerge true also didn't work as expected - it is removed here.

References:

- https://docs.renovatebot.com/presets-group/#groupallnonmajor
- https://stackoverflow.com/questions/66471226/renovate-combine-all-updates-to-one-branch-pr

See:

- https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/451
- https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/449
- https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/448
- https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/447
- https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/446
- https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/445
- https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/444
- https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/443
- https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/442
- https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/441

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -7,6 +7,12 @@
     "description": "Create a PR whenever there is a new major version",
     "matchUpdateTypes": ["major"]
   }],
+  "patch": {
+    "enabled": false
+  },
+  "minor": {
+    "enabled": false
+  },
   "ignorePaths": [
     "docs/**",
     "experimental/**"
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -3,16 +3,26 @@
   "extends": [
     "config:base"
   ],
-  "packageRules": [{
-    "description": "Create a PR whenever there is a new major version",
-    "matchUpdateTypes": ["major"]
-  }],
-  "patch": {
-    "enabled": false
-  },
-  "minor": {
-    "enabled": false
-  },
+  "packageRules": [
+    {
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -1,13 +1,18 @@
 {
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
```
anniefu, line 2: Is the problem with automerge just that a reviewer needs to approve?

To reduce noise, we may want to try batched updates instead: https://docs.renovatebot.com/getting-started/use-cases/#batched-updates. There may be some packages that never release a major version change because they don't have breaking changes, but are pushing out new features or fixes frequently.

grant: No, the problem is that we have too many PRs without significant updates.

Batched updates should also work well. I've updated as such.

anniefu (approved):

PR #457, opened 2022-04-20
fix(deps): Move `@types/express` to `dependencies`

koistya:
Otherwise, `req`, `res` vars have no members in VSCode auto-complete (other than `req.rawBody`).

Reference https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html


grant (approved):
Thanks!

PR #465, opened 2022-06-08
ci: test node 18

grant:
- Test Node 18.


grant:
We can do this later when we are working on this for GCF.

PR #497, opened 2022-12-27
fix: remove default fallback when target function does not exist

garethgeorge:
The function framework's current behavior is to load an exported function named 'function' if one exists as a fallback if the user's declared target function is not found. This behavior is counter intuitive and can hide misconfiguration from the user.

In this CL a breaking change is made to disable this fallback behavior in favor of failing fast so that the configuration error is highly visible to the user.

### DIFF
--- a/src/loader.ts
+++ b/src/loader.ts
@@ -121,7 +121,7 @@ export async function getUserFunction(
       return registeredFunction;
     }
 
-    let userFunction = functionTarget
+    const userFunction = functionTarget
       .split('.')
       .reduce((code, functionTargetPart) => {
         if (typeof code === 'undefined') {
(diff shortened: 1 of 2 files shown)


src/loader.ts
```diff
@@ -131,18 +131,12 @@ export async function getUserFunction(
...
-      if (functionModule.hasOwnProperty('function')) {
-        userFunction = functionModule['function'];
-      } else {
-        console.error(
-          `Function '${functionTarget}' is not defined in the provided ` +
-            'module.\nDid you specify the correct target function to execute?'
-        );
-        return null;
-      }
+      console.error(
```
anniefu, line 135: Does this impact the case where no `--target` flag or `FUNCTION_TARGET` env var is provided?

I think we want to still keep the default behavior documented in the README for that, where we look for a function named `function`

i.e. running just `functions-framework` starts up a function named `function` still

However if someone runs `functions-framework --target=myFn` and `myFn` does **not** exist we should error instead of starting up a function named `function`, because that's a conflict with the user's specifications.

garethgeorge: It shouldn't affect either case, the target / FUNCTION_TARGET [provides 'function' as a default value](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/src/options.ts#L86-L90) so when no value is provided it will fallback to the expected default behavior. 

We have some test coverage [here](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/test/options.ts#L50).

anniefu (approved):

anniefu (changes requested):
Actually, I'm not sure we want to release this as a breaking change. I think it's just a small bug fix rather than an API change and doesn't really warrant a major version increment.

Could you change the commit description to just `fix:`?

https://github.com/googleapis/release-please#how-should-i-write-my-commits

garethgeorge:
Done, I wasn't sure if we might want to prevent users from updating to the new behavior. I suspect very few customers are depending on this behavior as it is not part of the contract & is not very expected.

anniefu (approved):

PR #505, opened 2023-01-09
fix: make renovatebot's ignorePath and grouping of non-major updates work correctly

kenneth-rosario:
The renovate configs currently in use is not correctly ignoring the docs and experimental folders, and seems to not group non major updates together.

* The previous renovate config ```ignorePaths``` was not working due to a current issue with renovate itself where ```ignorePaths``` is being overwrited by the ```coinfig:base``` defaults. The suggested fix is to use ```globalExtends``` property. Discussion used: https://github.com/renovatebot/renovate/discussions/14793
* Changed config to use the ```group:allNonMajor``` extension which ended up grouping non-major versions correctly.

Tested by forking repo, changing versions manually, and then running a locally cloned renovate bot against my fork with the updated renovate config file.

Example of how non major grouping would look like:
Screenshot 2023-01-09 at 11 43 15 AM

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,34 +1,20 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": [
+  "globalExtends": [
     "config:base"
   ],
+  "extends": ["group:allNonMajor", "schedule:monthly"],
   "packageRules": [
     {
       "description": "Create a PR whenever there is a new major version",
       "matchUpdateTypes": [
         "major"
       ]
-    },
-    {
-      "description": "Create a PR grouping all non-major dependencies",
-      "matchPackagePatterns": [
-        "*"
-      ],
-      "matchUpdateTypes": [
-        "minor",
-        "patch"
-      ],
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -1,26 +1,15 @@
...
-      ],
-      "matchUpdateTypes": [
-        "minor",
-        "patch"
-      ],
-      "groupName": "all non-major dependencies",
-      "groupSlug": "all-minor-patch"
     }
   ],
   "ignorePaths": [
```
anniefu, line 15: Could you also make the schedule more frequent, like once a month, now that the PRs are going to be consolidated"

anniefu (approved):
Nice! This should make the PRs way more manageable and we can actually merge all the updates routinely now.

.github/renovate.json
kenneth-rosario: Sounds good, changed config to extend the schedule:monthly renovate preset

wyardley:
(side note, probably this should have been a non-triggering type like chore or ci / build vs. fix)?

PR #531, opened 2023-05-09
chore: Add NodeJS 18 and NodeJS 20 as test targets for github workflows

MontyCarter:


kenneth-rosario:
I think you will need to amend current commit to renam and force push to branch for conventionalcommit to not complain.

MontyCarter:
> I think you will need to amend current commit to renam and force push to branch for conventionalcommit to not complain.

Done.

kenneth-rosario (approved):

PR #545, opened 2023-06-15
fix: framework fails to shutdown gracefully when exit code is a string

garethgeorge:
Despite the Node.js docs indicating that the [exit code should always be a number](https://nodejs.org/docs/latest-v10.x/api/process.html#process_event_exit) it seems this is unchecked and, in fact, `process.exit` may be called with a string value.

Verified experimentally:
```
// test.js
const process = require("process");

process.on('exit', function(code) {
   console.log(typeof code);
   console.log(code);
});

process.exit('abc');
// output
string
abc
```

Some Cloud Function users report errors on shutdown when either their code or a dependency calls process.exit with a string value. To account for this the exit code check is updated to use string comparison to account for the possibility of a string exit code.

### DIFF
--- a/src/invoker.ts
+++ b/src/invoker.ts
@@ -100,7 +100,10 @@ export class ErrorHandler {
       sendCrashResponse({err, res: latestRes, callback: killInstance});
     });
 
-    process.on('exit', code => {
+    process.on('exit', (code: number | string) => {
+      if (typeof code === 'string') {
+        code = parseInt(code);
+      }
       sendCrashResponse({
         err: new Error(`Process exited with code ${code}`),
         res: latestRes,


garethgeorge:
No testing at the moment, it's unclear how we should approach testing this (beyond local exploratory testing). We don't have process lifecycle tests today and mocha doesn't support this type of test. 

I think we could possibly try to add a test that simulates emitting the event, I'm also fine with relying on exploratory tests and our conformance test coverage which should be exercising the full process lifecycle. WDYT?

src/invoker.ts
```diff
@@ -100,7 +100,10 @@ export class ErrorHandler {
       sendCrashResponse({err, res: latestRes, callback: killInstance});
     });
 
-    process.on('exit', code => {
+    process.on('exit', (code: number | string) => {
+      if (typeof code === 'string') {
+        code = parseInt(code);
```
guangyw, line 105: should we change how we compare(line 110, to avoid crashing when code is string), than try to parse the string?

guangyw: it doesn't look like the comparison will cause crash either. Do you know where in our code it's crashing?

garethgeorge: The comparison isn't actually what's crashing -- it compares successfully but gets the value wrong since "0" === 0 is false. We'd like it to still exit silently when "0" (string zero) is provided.

guangyw (approved):

PR #553, opened 2023-07-07
chore: make buildpack integ test run on pull request

kenneth-rosario:


.github/workflows/buildpack-integration-test.yml
```diff
@@ -4,6 +4,10 @@ on:
   push:
     branches:
       - master
+  pull_request:
+  # Runs every day on 12:00 AM PST
+  schedule:
+    - cron: "0 0 * * 1"
```
garethgeorge, line 10: nit: this is only on Mondays I think, for daily at midnight we'd want `0 0 * * *`

test/conformance/prerun.sh
```diff
@@ -17,5 +23,10 @@ npm pack --pack-destination $SCRIPT_DIR
...
 cd $SCRIPT_DIR
 
+# Modify package.json to include the desired node version
+tmp=$(mktemp)
+jq --arg version "$NODE_VERSION" '.engines.node = $version' package.json >"$tmp" && mv "$tmp" package.json
+
 npm install google-cloud-functions-framework-0.0.0.tgz
-cat package.json
\ No newline at end of file
+rm package-lock.json
```
garethgeorge, line 31: Why remove the `package-lock.json` ? When using npm wouldn't we typically expect customers to include this with their source code?

kenneth-rosario: Previous to this PR our buildpack integrations tests always used the latest version of node instead of the one specified in the builder runtime. See https://github.com/GoogleCloudPlatform/functions-framework-nodejs/actions/runs/5393557655/jobs/9793478045#step:6:346 

The package-lock.json is generated by the prerun script using a different node version than the one the function requires.

I was getting following error on node 14 and node 12:
https://github.com/GoogleCloudPlatform/functions-framework-nodejs/actions/runs/5490281343/jobs/10005532209#step:6:388

I was getting the same error when deploying when the package-lock.json was created using a different node version on nodejs12 and nodejs14.

kenneth-rosario: @garethgeorge moved the runtime version pinning into the conformance client as a new flag.

garethgeorge (approved):

PR #587, opened 2024-01-14
update readme.md to include ESM example

blazing234:
update readme.md to include ESM example

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ npm install @google-cloud/functions-framework
 
 1. Create an `index.js` file with the following contents:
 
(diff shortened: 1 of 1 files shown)


blazing234:
Signed CLA

blazing234:
@kenneth-rosario Would this type of PR be able to be approved? I want to edit the cloud docs in general to include more ESM examples

matthewrobertson:
We have a guide in docs that described how to use ESM: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/main/docs/esm.md

I don't think we want to pull this into the main README as we want to keep in minimal.

PR #592, opened 2024-02-22
feat: add execution id support

liuyunnnn:
Add execution id support.

When enabled with `--enable-execution-id` or set `ENABLE_EXECUTION_ID=True` in environment variables, execution id will be added to log.

### DIFF
--- a/src/execution_context.ts
+++ b/src/execution_context.ts
@@ -1,0 +1,55 @@
+import {AsyncLocalStorage} from 'node:async_hooks';
+import {Request, Response, NextFunction} from 'express';
+
+export const asyncLocalStorage = new AsyncLocalStorage();
+
+export interface ExeuctionContext {
+  'logging.googleapis.com/labels': {
+    executionId: string;
+  };
+  'logging.googleapis.com/trace'?: string;
+  'logging.googleapis.com/spanId'?: string;
+}
+
+export const executionContextMiddleware = (
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  let executionId = req.header('function-execution-id');
+  if (!executionId || executionId.length !== 12) {
+    executionId = generateExecutionId();
+  }
+
+  let traceId, spanId, options;
+  const cloudTraceContext = req.header('X-Cloud-Trace-Context');
+  const regex = /^(?<traceId>\w+)\/(?<spanId>\d+);o=(?<options>.+)$/;
+  if (cloudTraceContext) {
+    const match = cloudTraceContext.match(regex);
+    if (match?.groups) {
+      ({traceId, spanId, options} = match.groups);
+    }
+  }
+
+  const executionContext = <ExeuctionContext>{
+    'logging.googleapis.com/labels': {
+      executionId: executionId,
+    },
+    'logging.googleapis.com/trace': traceId,
+    'logging.googleapis.com/spanId': spanId,
+  };
+
+  asyncLocalStorage.run(executionContext, () => {
+    next();
+  });
+};
+
+export function getCurrentContext() {
+  return asyncLocalStorage.getStore();
+}
+
+function generateExecutionId() {
+  const timestampPart = Date.now().toString(36).slice(-6);
+  const randomPart = Math.random().toString(36).slice(-6);
+  return timestampPart + randomPart;
+}
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -14,6 +14,8 @@
 
 import * as express from 'express';
 import {FUNCTION_STATUS_HEADER_FIELD} from './types';
+import {getCurrentContext, ExeuctionContext} from './execution_context';
+import {Buffer} from 'buffer';
 
 /**
  * Logs an error message and sends back an error response to the incoming
@@ -58,5 +60,81 @@ export function sendCrashResponse({
   }
   if (callback) {
     callback();
+  }
+}
+
+export function loggingHandlerAddExecutionContext() {
+  const originalStdoutWrite = process.stdout.write;
+  process.stdout.write = (data, ...args) => {
+    const {encoding, cb} = splitArgs(args);
+    const modifiedData = getModifiedData(data, encoding);
+    return originalStdoutWrite.apply(process.stdout, [modifiedData, cb]);
+  };
+
+  const originalStderrWrite = process.stderr.write;
+  process.stderr.write = (data, ...args) => {
+    const {encoding, cb} = splitArgs(args);
+    const modifiedData = getModifiedData(data, encoding, true);
+    return originalStderrWrite.apply(process.stderr, [modifiedData, cb]);
+  };
+}
+
+export function splitArgs(args: any) {
+  let encoding, cb;
+  if (
+    args.length > 0 &&
+    (Buffer.isEncoding(args[0]) || typeof args[0] === 'undefined')
+  ) {
+    encoding = args[0];
+    args.shift();
+  }
+  if (args.length > 0 && typeof args[0] === 'function') {
+    cb = args[0];
+  }
+  return {encoding: encoding, cb: cb};
+}
+
+export function getModifiedData(
+  data: Uint8Array | string,
+  encoding?: BufferEncoding,
+  stderr: boolean = false
+) {
+  let modifiedData = data;
+  if (getCurrentContext()) {
+    const exeuctionContext = getCurrentContext() as ExeuctionContext;
+    const {isJSON, processdData} = processData(data, encoding);
+    let dataWithContext;
+    if (isJSON) {
+      dataWithContext = {...exeuctionContext, ...processdData};
+    } else {
+      dataWithContext = {...exeuctionContext, message: processdData};
+    }
+    if (stderr) {
+      dataWithContext['severity'] = 'ERROR';
+    }
+
+    modifiedData = JSON.stringify(dataWithContext) + '\n';
+  }
+  return modifiedData;
+}
+
+function processData(data: Uint8Array | string, encoding?: BufferEncoding) {
+  let decodedData;
+  try {
+    if (data instanceof Uint8Array) {
+      const textDecoder = new TextDecoder('utf-8');
+      decodedData = textDecoder.decode(data);
+    } else {
+      decodedData = Buffer.from(data, encoding).toString('utf-8');
+    }
+  } catch (e) {
+    // Failed to decode, treat it as simple text.
+    return {isJSON: false, processdData: data};
+  }
+
+  try {
+    return {isJSON: true, processdData: JSON.parse(decodedData)};
+  } catch (e) {
+    return {isJSON: false, processdData: decodedData};
   }
 }
--- a/src/options.ts
+++ b/src/options.ts
@@ -47,6 +47,10 @@ export interface FrameworkOptions {
    * Whether or not the --help CLI flag was provided.
    */
   printHelp: boolean;
+  /**
+   * Whether or not to enable execution id support.
+   */
+  enableExecutionId: boolean;
 }
 
 /**
@@ -107,6 +111,17 @@ const SignatureOption = new ConfigurableOption(
     );
   }
 );
+const EnableExecutionIdOption = new ConfigurableOption(
+  'enable-execution-id',
+  'ENABLE_EXECUTION_ID',
+  false,
+  x => {
+    return (
+      (typeof x === 'boolean' && x) ||
+      (typeof x === 'string' && x.toLowerCase() === 'true')
+    );
+  }
+);
 
 export const helpText = `Example usage:
   functions-framework --target=helloWorld --port=8080
@@ -131,6 +146,7 @@ export const parseOptions = (
       SignatureOption.cliOption,
       SourceLocationOption.cliOption,
     ],
+    boolean: [EnableExecutionIdOption.cliOption,],
   });
   return {
     port: PortOption.parse(argv, envVars),
@@ -138,5 +154,6 @@ export const parseOptions = (
     sourceLocation: SourceLocationOption.parse(argv, envVars),
     signatureType: SignatureOption.parse(argv, envVars),
     printHelp: cliArgs[2] === '-h' || cliArgs[2] === '--help',
+    enableExecutionId: EnableExecutionIdOption.parse(argv, envVars),
   };
 };
--- a/README.md
+++ b/README.md
@@ -190,6 +190,7 @@ ignored.
 | `--target`         | `FUNCTION_TARGET`         | The name of the exported function to be invoked in response to requests. Default: `function`                                                                                                                     |
 | `--signature-type` | `FUNCTION_SIGNATURE_TYPE` | The signature used when writing your function. Controls unmarshalling rules and determines which arguments are used to invoke your function. Default: `http`; accepted values: `http` or `event` or `cloudevent` |
 | `--source`         | `FUNCTION_SOURCE`         | The path to the directory of your function. Default: `cwd` (the current working directory)                                                                                                                       |
+| `--enable-execution-id`         | `ENABLE_EXECUTION_ID`         | Whether or not to enable execution id support. When enable, exeuction id will be added to log.                                                                                                                     |
 
 You can set command-line flags in your `package.json` via the `start` script.
 For example:
--- a/src/main.ts
+++ b/src/main.ts
@@ -20,6 +20,7 @@ import {getUserFunction} from './loader';
 import {ErrorHandler} from './invoker';
 import {getServer} from './server';
 import {parseOptions, helpText, OptionsError} from './options';
+import {loggingHandlerAddExecutionContext} from './logger';
 
 /**
  * Main entrypoint for the functions framework that loads the user's function
@@ -27,6 +28,8 @@ import {parseOptions, helpText, OptionsError} from './options';
  */
 export const main = async () => {
   try {
+    loggingHandlerAddExecutionContext();
+
     const options = parseOptions();
 
     if (options.printHelp) {
--- a/src/main.ts
+++ b/src/main.ts
@@ -28,14 +28,17 @@ import {loggingHandlerAddExecutionContext} from './logger';
  */
 export const main = async () => {
   try {
-    loggingHandlerAddExecutionContext();
-
     const options = parseOptions();
 
     if (options.printHelp) {
       console.error(helpText);
       return;
     }
+
+    if (options.enableExecutionId) {
+      loggingHandlerAddExecutionContext();
+    }
+
     const loadedFunction = await getUserFunction(
       options.sourceLocation,
       options.target,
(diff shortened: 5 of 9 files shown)


src/logger.ts
```diff
@@ -60,3 +62,79 @@ export function sendCrashResponse({
...
+    modifiedData = JSON.stringify(dataWithContext) + '\n';
+  }
+  return modifiedData;
+}
+
+function processData(data: Uint8Array | string, encoding?: BufferEncoding) {
+  let decodedData;
+  try {
+    if (data instanceof Uint8Array) {
+      const textDecoder = new TextDecoder('utf-8');
```
kenneth-rosario, line 125: I wonder if we can avoid the use of TextDecoder all together. Would the answer 4 from top to down with 35 upvotes of https://stackoverflow.com/questions/8936984/uint8array-to-string-in-javascript help?

liuyunnnn: Done.

src/execution_context.ts
```diff
@@ -0,0 +1,69 @@
...
+  'logging.googleapis.com/spanId'?: string;
+}
+
+function generateExecutionId() {
+  const timestampPart = Date.now().toString(36).slice(-6);
+  const randomPart = Math.random().toString(36).slice(-6);
+  return timestampPart + randomPart;
+}
+
+let asyncLocalStorage: any;
```
kenneth-rosario, line 19: Can we do `AsyncLocalStorage | undefined` as the type?

src/logger.ts
```diff
@@ -60,3 +62,78 @@ export function sendCrashResponse({
...
+
+  const originalStderrWrite = process.stderr.write;
+  process.stderr.write = (data, ...args) => {
+    const {encoding, cb} = splitArgs(args);
+    const modifiedData = getModifiedData(data, encoding, true);
+    return originalStderrWrite.apply(process.stderr, [modifiedData, cb]);
+  };
+}
+
+export function splitArgs(args: any) {
```
kenneth-rosario, line 82: nit: any[] type since it looks like it is an array

src/execution_context.ts
```diff
@@ -0,0 +1,69 @@
...
+  next: NextFunction
+) {
+  if (semver.gte(process.versions.node, requriedNodeJsVersion)) {
+    const asyncHooks = await import('node:async_hooks');
+    if (!asyncLocalStorage) {
+      asyncLocalStorage = new asyncHooks.AsyncLocalStorage();
+    }
+
+    let executionId = req.header('function-execution-id');
+    if (!executionId || executionId.length !== 12) {
```
kenneth-rosario, line 33: Lets move the 12 to a constant. EXECUTION_ID_LENGTH

src/execution_context.ts
```diff
@@ -0,0 +1,69 @@
...
+      executionId = generateExecutionId();
+    }
+
+    let traceId, spanId, options;
+    const cloudTraceContext = req.header('X-Cloud-Trace-Context');
+    const regex = /^(?<traceId>\w+)\/(?<spanId>\d+);o=(?<options>.+)$/;
```
kenneth-rosario, line 39: Lets move the regex to a constant. TRACE_CONTEXT_PATTERN

src/execution_context.ts
```diff
@@ -0,0 +1,69 @@
...
+      asyncLocalStorage = new asyncHooks.AsyncLocalStorage();
+    }
+
+    let executionId = req.header('function-execution-id');
+    if (!executionId || executionId.length !== 12) {
+      executionId = generateExecutionId();
+    }
+
+    let traceId, spanId, options;
+    const cloudTraceContext = req.header('X-Cloud-Trace-Context');
```
kenneth-rosario, line 38: Lets move header key to a constant. TRACE_CONTEXT_HEADER_KEY.

src/execution_context.ts
```diff
@@ -0,0 +1,69 @@
...
+    const regex = /^(?<traceId>\w+)\/(?<spanId>\d+);o=(?<options>.+)$/;
+    if (cloudTraceContext) {
+      const match = cloudTraceContext.match(regex);
+      if (match?.groups) {
+        ({traceId, spanId, options} = match.groups);
+      }
+    }
+
+    const executionContext = <ExeuctionContext>{
+      'logging.googleapis.com/labels': {
```
kenneth-rosario, line 48: We should move this keys to a constant and use them throughout the codebase.

EXECUTION_CONTEXT_LABELS_KEY
EXECUTION_CONTEXT_TRACE_KEY
EXECUTION_CONTEXT_SPAN_ID_KEY

src/execution_context.ts
liuyunnnn: Var `asyncLocalStorage` is declared in the scope of module so that it can be shared between function `executionContextMiddleware` and `getCurrentContext`, while `AsyncLocalStorage` is imported dynamically in function `executionContextMiddleware`. I'm new to NodeJS/TypeScript so not pretty sure how to work with dynamic import and if it's desired to have a module variable like this. Wondering if you have any suggestions on a better structure.

kenneth-rosario: I am fairly new as well
@matthewrobertson any recommendations on this?

kenneth-rosario:
I think you need to run `npm run docs`

src/execution_context.ts
```diff
@@ -0,0 +1,76 @@
...
+    executionId = generateExecutionId();
+  }
+
+  let traceId, spanId, options;
+  const cloudTraceContext = req.header(TRACE_CONTEXT_HEADER_KEY);
+  const regex = TRACE_CONTEXT_PATTERN;
+  if (cloudTraceContext) {
+    const match = cloudTraceContext.match(regex);
+    if (match?.groups) {
+      ({traceId, spanId, options} = match.groups);
```
kenneth-rosario, line 52: nit: can we do `const {traceId, spanId} = match.groups`

src/logger.ts
```diff
@@ -60,3 +66,120 @@ export function sendCrashResponse({
...
+  }
+  const {isJSON, processdData} = processData(data, encoding);
+  let dataWithContext;
+  if (isJSON) {
+    dataWithContext = getJSONWithContext(processdData, currentContext);
+  } else {
+    dataWithContext = getTextWithContext(processdData, currentContext);
+  }
+  if (stderr) {
+    dataWithContext['severity'] = 'ERROR';
```
kenneth-rosario, line 136: Move severity to constant

src/options.ts
```diff
@@ -108,6 +113,27 @@ const SignatureOption = new ConfigurableOption(
   }
 );
 
+export const requriedNodeJsVersion = '13.0.0';
+function enableExecutionId(envVars: NodeJS.ProcessEnv): boolean {
+  const logExecutionID = envVars['LOG_EXECUTION_ID'];
```
kenneth-rosario, line 118: move env var to constant

kenneth-rosario (commented):
Can we create integration test for this or would it be too complicated? similar to test/integration/http.ts . Not sure if we could redirect stdout to some sort of pipe and be able inspect the contents or make test function dump stdout on res.body. Let me know what you think.

src/logger.ts
```diff
@@ -60,3 +66,120 @@ export function sendCrashResponse({
     callback();
   }
 }
+
+export function loggingHandlerAddExecutionContext() {
```
nifflets, line 70: this is great! Were you able to do any end to end testing on this to see the final output logs?

src/execution_context.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export function getCurrentContext(): ExecutionContext | undefined {
+  if (!asyncLocalStorage) {
+    return undefined;
+  }
+  return asyncLocalStorage.getStore();
+}
+
+export const getCurrentExecutionId = (): string | undefined => {
```
nifflets, line 74: Can you add tests for this? Including calling it outside of a request context

src/logger.ts
```diff
@@ -60,3 +66,120 @@ export function sendCrashResponse({
...
+  interceptStdoutWrite();
```
nifflets, line 71: Also, we weren't planning to enable the stdout/stderr interception quite yet, but I'm okay with it being gated behind an env var. cc @matthewrobertson for additional opinion

src/execution_context.ts
liuyunnnn: traceId and spanId are defined previously, out of the scope of retrieving trace them header and matching regex. Personally I feel re-declaring them would complicate the situation.

src/execution_context.ts
liuyunnnn: It's already in tests. The store is only accessible to "asynchronous operations created within the callback" of `run()`, so they are checked in `next()` function.

I just added an additional assert that calling it outside of next() will return undefined.

src/logger.ts
liuyunnnn: Yes. For example, by`console.log('Hello log!');`, it logs
`{"message":"Hello log!\n","logging.googleapis.com/labels":{"execution_id":"5a80ltr2xe5s"}}`.

However it doesn't work perfectly. For example, `console.log()` always extends a `\n` at the end of the string, which will go to the `message` field as well.

 Similarly, for string with format, like `throw Error`, it outputs `{"message":"[Function: Error] {\n  stackTraceLimit: \u001b[33m10\u001b[39m,\n  prepareStackTrace: \u001b[90mundefined\u001b[39m,\n  domainThrown: \u001b[33mtrue\u001b[39m\n}\n","logging.googleapis.com/labels":{"execution_id":"5ae4layspp8k"},"severity":"ERROR"}`,
comparing to original log in console and [cloud log](https://screenshot.googleplex.com/BaNwXQr9wAuCJ73)
image


Personally I think it's ok as it conveys the same message. WDYT?

src/execution_context.ts
```diff
@@ -0,0 +1,76 @@
...
+    next();
+    return;
+  }
+  const asyncHooks = await import('node:async_hooks');
+  if (!asyncLocalStorage) {
+    asyncLocalStorage = new asyncHooks.AsyncLocalStorage();
+  }
+
+  let executionId = req.header(FUNCTION_EXECUTION_ID_HEADER_KEY);
+  if (!executionId || executionId.length !== EXECUTION_ID_LENGTH) {
```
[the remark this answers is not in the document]
liuyunnnn: Since user can provide execution id, I will assume no check is needed.

src/logger.ts
```diff
@@ -60,3 +66,120 @@ export function sendCrashResponse({
...
+  let decodedData;
+  try {
+    if (data instanceof Uint8Array) {
+      decodedData = Buffer.from(data.buffer).toString();
+    } else {
+      decodedData = Buffer.from(data, encoding).toString();
+    }
+  } catch (e) {
+    // Failed to decode, treat it as simple text.
+    return {isJSON: false, processdData: data};
```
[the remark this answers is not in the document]
liuyunnnn: Thanks for catching! I'm spoiled by cider-v's typo detection :(

src/execution_context.ts
```diff
@@ -0,0 +1,69 @@
+import * as semver from 'semver';
+import {Request, Response, NextFunction} from 'express';
+import {requriedNodeJsVersion} from './options';
+
+export interface ExeuctionContext {
```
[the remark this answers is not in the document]
liuyunnnn: This typo is already fixed in previous commit.

liuyunnnn:
> Can we create integration test for this or would it be too complicated? similar to test/integration/http.ts . Not sure if we could redirect stdout to some sort of pipe and be able inspect the contents or make test function dump stdout on res.body. Let me know what you think.

I haven't figure out how to do that yet. All the ways I found are overriding console.log or process.stdout.write, which is the function we monkey patch in this PR.

src/execution_context.ts
```diff
@@ -0,0 +1,74 @@
...
+    const match = cloudTraceContext.match(TRACE_CONTEXT_PATTERN);
+    if (match?.groups) {
+      ({traceId, spanId, options} = match.groups);
+    }
+  }
+
+  const executionContext = <ExecutionContext>{
+    executionId: executionId,
+    traceId: traceId,
+    spanId: spanId,
```
matthewrobertson, line 57: there is shorthand for this if the keys and variables are the same. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer

src/execution_context.ts
```diff
@@ -0,0 +1,74 @@
...
+  let traceId, spanId, options;
+  const cloudTraceContext = req.header(TRACE_CONTEXT_HEADER_KEY);
+  if (cloudTraceContext) {
+    const match = cloudTraceContext.match(TRACE_CONTEXT_PATTERN);
+    if (match?.groups) {
+      ({traceId, spanId, options} = match.groups);
+    }
+  }
+
+  const executionContext = <ExecutionContext>{
```
matthewrobertson, line 54: I don't think you need to cast this.

src/execution_context.ts
```diff
@@ -0,0 +1,74 @@
...
+  if (!executionId) {
+    executionId = generateExecutionId();
+  }
+
+  let traceId, spanId, options;
+  const cloudTraceContext = req.header(TRACE_CONTEXT_HEADER_KEY);
+  if (cloudTraceContext) {
+    const match = cloudTraceContext.match(TRACE_CONTEXT_PATTERN);
+    if (match?.groups) {
+      ({traceId, spanId, options} = match.groups);
```
matthewrobertson, line 50: you don't use options so don't assign it.

src/execution_context.ts
matthewrobertson: the typing for this stuff will get a log simpler if you split out the logic for parsing the trace context into its own function.

src/execution_context.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+let asyncLocalStorage: any;
+
+export async function executionContextMiddleware(
+  req: Request,
+  res: Response,
+  next: NextFunction
+) {
+  if (semver.lt(process.versions.node, requiredNodeJsVersion)) {
```
matthewrobertson, line 30: I think you need to split out the logic for parsing context from the logic for configuring the asyncLocalStorage into separate middlewares.

src/execution_context.ts
liuyunnnn: Explained offline. asyncHooks has to be imported dynamically.

src/logger.ts
```diff
@@ -60,3 +67,120 @@ export function sendCrashResponse({
...
+    encoding = args[0];
+    args.shift();
+  }
+  if (args.length > 0 && typeof args[0] === 'function') {
+    cb = args[0];
+  }
+  return {encoding: encoding, cb: cb};
+}
+
+export function getModifiedData(
```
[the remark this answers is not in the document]
liuyunnnn: It's used and test in unittest. The `loggingHandlerAddExecutionContext` function is hard to unittest as it's monkeypatched `process.stdout.write`. To test it, it needs to redirect stdout to somewhere and check the contents, but currently all the methods I found to validate log is to monkey patch `process.stdout.write` or `console.log`, which doesn't really work in our case.

src/options.ts
```diff
@@ -138,5 +165,6 @@ export const parseOptions = (
     sourceLocation: SourceLocationOption.parse(argv, envVars),
     signatureType: SignatureOption.parse(argv, envVars),
     printHelp: cliArgs[2] === '-h' || cliArgs[2] === '--help',
+    enableExecutionId: enableExecutionId(envVars),
```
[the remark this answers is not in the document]
liuyunnnn: In current design, we don't provide a cli flag to enable log execution id, but only allow enable it by env var. Seems to me `ConfigurableOption` provides both.

src/execution_context.ts
```diff
@@ -0,0 +1,74 @@
...
+  spanId?: string;
+}
+
+const TRACE_CONTEXT_PATTERN =
+  /^(?<traceId>\w+)\/(?<spanId>\d+);o=(?<options>.+)$/;
+
+function generateExecutionId() {
+  const timestampPart = Date.now().toString(36).slice(-6);
+  const randomPart = Math.random().toString(36).slice(-6);
+  return timestampPart + randomPart;
```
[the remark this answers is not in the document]
liuyunnnn: Tried but all CI/CD tests with nodejs version < 16 failed. e.g. https://github.com/GoogleCloudPlatform/functions-framework-nodejs/actions/runs/8136501060/job/22232943740?pr=592

kenneth-rosario (approved):
LGTM, please wait on [nifflets](https://github.com/nifflets)'s approval before merging

src/options.ts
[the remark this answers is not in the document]
matthewrobertson: I don't think there is a problem with creating a CLI flag for this. Either way you can still make a `ConfigurableOption` just don't pass it to `minimist` here: https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/def93dae74a53204f5db51842592583fe6f746ce/src/options.ts#L127

src/async_local_storage.ts
```diff
@@ -0,0 +1,52 @@
...
+import {NextFunction} from 'express';
+import {requiredNodeJsVersionForLogExecutionID} from './options';
+
+export interface ExecutionContext {
+  executionId?: string;
+  traceId?: string;
+  spanId?: string;
+}
+
+let asyncLocalStorage: any;
```
matthewrobertson, line 12: please fix: this should be probably be typed as `AsyncLocalStorage<ExecutionContext> | undefined`

docs/generated/api.md
```diff
@@ -58,6 +58,9 @@ export interface EventFunctionWithCallback {
     (data: {}, context: Context, callback: Function): any;
 }
 
+// @public
+export const getExecutionId: () => string | undefined;
```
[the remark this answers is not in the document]
matthewrobertson: Let's delete this for now. I will try to introduce it a context object in a subsequent CL.

src/async_local_storage.ts
```diff
@@ -0,0 +1,52 @@
...
+  next: NextFunction
+) {
+  if (
+    semver.lt(process.versions.node, requiredNodeJsVersionForLogExecutionID)
+  ) {
+    // Skip for unsupported Node.js version.
+    next();
+    return;
+  }
+  const asyncHooks = await import('node:async_hooks');
```
matthewrobertson, line 26: looks like this import should move inside the `if` below

src/function_registry.ts
```diff
@@ -65,6 +66,14 @@ export const isValidFunctionName = (functionName: string): boolean => {
   return regex.test(functionName);
 };
 
+/**
+ * Gets the request-specific execution id
+ * @public
+ */
+export const getExecutionId = (): string | undefined => {
```
matthewrobertson, line 73: As above. Let's drop this function for now.

src/index.ts
```diff
@@ -20,4 +20,4 @@ export * from './functions';
 /**
  * @public
  */
-export {http, cloudEvent, typed} from './function_registry';
+export {http, cloudEvent, typed, getExecutionId} from './function_registry';
```
matthewrobertson, line 23: ditto: remove

matthewrobertson (approved):
Overall looks good. Lets drop the global `getExecutionId` function from the API for now.

src/logger.ts
liuyunnnn: Discussed offline we're okay with it.

src/async_local_storage.ts
matthewrobertson: Oh you can import the type. Let me find a link.

matthewrobertson (approved):

kenneth-rosario (approved):

nifflets (approved):

PR #603, opened 2024-04-26
chore: udpate ts-doc annotations

matthewrobertson:
Fix a bunch of ts-doc formatting issues that were spewing warnings.


HKWinterhalter (approved):
Thanks for the fix

PR #646, opened 2024-10-18
feat: added new option to change the server host 

lukascaska:
I am using this package to test my function and noticed something curious: 
I wasn’t able to access my function from inside a Docker container. I was using http://host.docker.internal:3000 and it returned a "connection refused" error. I ran a test trying to access my frontend, which was also running locally (but with the host set to 0.0.0.0), and I was able to access it normally.

To work around the issue, I ran my container using "network_mode": "host" (in the docker-compose file), so everything was on the same network and it worked perfectly. However, to avoid this adjustment and make everything simpler, I decided to suggest this change.

1. Run a function locally
2. Create any Ubuntu container
3. Execute inside the container and try to access your application using http://host.docker.internal:3000 (should return connection refused)
4. As a test, try accessing any other server where you have the option to set 0.0.0.0


simoneb (approved):
LGTM, useful addition

PR #660, opened 2024-12-10
chore(deps): upgrade all dependencies

matthewrobertson:
Updated all dependencies and fixed resulting broken tests. 

Unfortunately, we are still stuck on some older versions of packages until we completely drop support for nodejs10 and nodejs12 in 2025.


HKWinterhalter (approved):

package.json
```diff
@@ -19,14 +19,14 @@
     }
   },
   "dependencies": {
-    "@types/express": "4.17.21",
+    "@types/express": "^5.0.0",
```
zymotik, line 22: I see Express v4 referenced in dependencies, so the v5 types here are causing compatibility issues when used in other libraries. See issue #673.

Should types also be moved to `devDependencies`?

PR #680, opened 2025-03-27
chore: nodejs versions in github workflows

matthewrobertson:
Updates all our Github workflows to drop versions 10, 12, 14, and 16 which are all deprecated. And then start including 22.


jama22:
Makes sense as per lifecycle https://cloud.google.com/functions/docs/runtime-support

jama22 (approved):

PR #681, opened 2025-03-27
chore!: drop support for old nodejs versions

matthewrobertson:
This commit drops support for all Node.js versions < 18.0.0 as they are no longer supported by Cloud Run functions: https://cloud.google.com/functions/docs/runtime-support

chore!: upgrade all dependencies

fix!: delete the undocumented "typed" function signature


.github/workflows/conformance.yml
maemayve, line 1: Undocumented doesn't necessarily mean unused, do we know that this is unused?

src/function_registry.ts
```diff
@@ -16,30 +16,28 @@ import {
...
  * Singleton map to hold the registered functions
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const registrationContainer = new Map<string, RegisteredFunction<any, any>>();
+const registrationContainer = new Map<string, RegisteredFunction<any>>();
 
 /**
  * Helper method to store a registered function in the registration container
  */
 const register = <T = unknown, U = unknown>(
```
maemayve, line 37: Seems like U = unknown can be pulled

maemayve (approved):
My primary question is what evidence is there that supports the removal of typed function

jama22 (approved):
lgtm, ok to drop support for older nodejs as per runtime policies

caugner:
@matthewrobertson For future Node.js bumps, I would suggest to mention the dropped Node.js version in the title, so it'll show up in the release notes / changelog. 🙏

PR #764, opened 2026-06-06
fix: add typesVersions so testing module resolves under classic TypeScript moduleResolution

### LINKED ISSUE #594: Supertest: getTestServer Not Found
Hello and to whom it may concern.

I am trying to work with the implementation of [getTestServer in Supertest documentation](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/main/docs/testing-functions.md#integration-testing-with-supertest), and when I try to instantiate my test server I get this error:

Screenshot 2024-03-25 at 9 58 27 AM

@google-cloud/functions-framework is installed as a dependency of this project so my assumption would have been that this should be available in the import.

I've Googled a few things and I've tried updating my tsconfig, but nothing seems to work. 

Thanks ahead.

64johnlee:
## Problem

The `./testing` subpath export is declared in `exports` correctly, but TypeScript's default `moduleResolution: "node"` does not read the `exports` field in `package.json`. This means projects using the default (or older) TypeScript configuration get a type error when importing the testing utilities:

```ts
import { getTestServer } from '@google-cloud/functions-framework/testing';
// Error: Module '@google-cloud/functions-framework/testing' not found
```

Both `getTestServer` and `getFunction` are affected.

## Root Cause

TypeScript only respects `exports` subpath maps when `moduleResolution` is set to `"node16"`, `"nodenext"`, or `"bundler"`. Most projects still use `"node"` (the classic resolver), which only looks at `main` and `types` at the package root.

## Fix

Add a `typesVersions` field, which is the standard way to declare subpath type mappings for all TypeScript versions and moduleResolution modes:

```json
"typesVersions": {
  "*": {
    "testing": ["./build/src/testing.d.ts"]
  }
}
```

This is additive — projects already using `node16`/`bundler` continue using the `exports` map; projects using classic `node` resolution now also work.

## Verification

After this change, the following import works without `moduleResolution: "node16"`:

```ts
import { getTestServer, getFunction } from '@google-cloud/functions-framework/testing';
```

Fixes #594

🤖 Generated with Claude Code

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,13 @@
       "default": "./build/src/testing.js"
     }
   },
+  "typesVersions": {
+    "*": {
+      "testing": [
+        "./build/src/testing.d.ts"
+      ]
+    }
+  },
   "dependencies": {
     "@types/express": "^5.0.0",
     "body-parser": "^2.2.2",


taeold:
@64johnlee great contribution. can you sign the CLA if you are interesting in seeing this PR be merged?

taeold:
I created a manual patch for this PR (coauthored by the OP). Thanks for your contribution!

64johnlee:
Thanks @taeold — appreciate you turning this into a clean patch and keeping the co-author credit. Glad it helps the classic-resolution users.

PR #793, opened 2026-08-08
feat(testing): configure ignored routes in test server

### LINKED ISSUE #778: Feature request: allow overriding `ignoredRoutes` in `getTestServer()`
`getTestServer()` hard-codes `ignoredRoutes: null`, which re-enables the default `/favicon.ico` and `/robots.txt` 404 short-circuit for HTTP functions. There's no way for a test to reproduce a production deployment that disables it via `--ignored-routes ""`, so functions that legitimately serve those paths can't be tested through `getTestServer()`.

Proposal: accept an optional options bag that overrides the `getServer` options (starting with `ignoredRoutes`), defaulting to today's behavior:

```ts
const server = getTestServer('myFn', {ignoredRoutes: ''});
```

Backward compatible since the parameter is optional. `ignoredRoutes` is the immediate need; a general options bag would also cover `timeoutMilliseconds` etc.

Version: `@google-cloud/functions-framework` 5.0.5.

hugosmoreira:
## Summary

- add an optional `TestServerOptions` bag to `getTestServer`
- pass `ignoredRoutes` through while preserving the existing default behavior
- document disabling the default favicon and robots.txt short-circuit
- test defaults, both ignored paths, empty/whitespace and custom overrides, error propagation, and the installed public `/testing` entry point

Fixes #778

## Validation

Local verification on Windows with Node 22.22.1 (generated API docs checked separately with Node 24.18.0):

- `npm test -- --reporter dot`: 126 passing, including 24 HTTP integration cases and two pack-and-install tests
- identical 25-case HTTP contract: 17 passing / 8 expected failures on the unchanged parent, all 25 passing with this feature
- negative control in an isolated checkout: bypassing the `ignoredRoutes` pass-through causes eight HTTP cases and the new installed-package test to fail as expected
- packed CommonJS and ESM runtime consumers pass
- exported `TestServerOptions` declaration checks pass with strict TypeScript 5.8.2, both Node16 and legacy Node10 module resolution, without `skipLibCheck`
- `npm run check`, compilation, changed-file ESLint/Prettier, and `git diff --check` pass; lint has one unchanged unused-generic warning also present on the parent
- `npm run docs` passes and generated documentation is unchanged
- all six non-Docker conformance groups pass using the workflow's pinned v1.8.3 client: legacy HTTP/event/CloudEvent, declarative HTTP/CloudEvent, and HTTP concurrency

Conformance used the same framework CLI directly on Windows to avoid an npm wrapper process; it was not a hosted Linux workflow run. The hosted OS/Node matrix and buildpack checks remain pending; the seven failure-labelled verification workflows on the previous revision expired awaiting approval without executing jobs.


hugosmoreira:
I've added regression coverage for both ignored routes, default/empty/custom options, error propagation, and the installed public `/testing` entry point. This follow-up changes tests only.

The full suite passes locally: 126 tests on Windows with Node 22.22.1. The new cases fail as expected when the option pass-through is bypassed. Packed CommonJS/ESM consumers, strict TypeScript declaration checks, lint/build/docs, and all six non-Docker conformance groups also pass. I've updated the PR description with the verification details and environments.

The seven failure-labelled verification workflow runs on the previous revision expired awaiting approval and had no jobs. Could a maintainer approve/rerun the checks for this update? The hosted OS/Node matrix and buildpack checks still need to run.

